Issue #113, opened 2019-04-02
Exception while describing TAP service with security methods


cbanek:
Here's what I'm trying to run:
```
import pyvo

service = pyvo.dal.TAPService('https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/tap')
service.describe()
```

Here's the callstack:
```
Traceback (most recent call last):
  File "test.py", line 4, in <module>
    service.describe()
  File "/home/cbanek/pyvo/pyvo/dal/tap.py", line 324, in describe
    self.capabilities
  File "/home/cbanek/pyvo-env/lib/python3.6/site-packages/astropy/utils/decorators.py", line 744, in __get__
    val = self.fget(obj)
  File "/home/cbanek/pyvo/pyvo/dal/vosi.py", line 91, in capabilities
    return vosi.parse_capabilities(self._capabilities().read)
  File "/home/cbanek/pyvo/pyvo/io/vosi/endpoint.py", line 120, in parse_capabilities
    config=config, pos=(1, 1)).parse(iterator, config)
  File "/home/cbanek/pyvo/pyvo/io/vosi/endpoint.py", line 333, in parse
    super(CapabilitiesFile, self).parse(iterator, config)
  File "/home/cbanek/pyvo/pyvo/utils/xml/elements.py", line 288, in parse
    iterator, tag, data, config, pos)
  File "/home/cbanek/pyvo/pyvo/io/vosi/endpoint.py", line 320, in capabilities
    capability.parse(iterator, config)
  File "/home/cbanek/pyvo/pyvo/utils/xml/elements.py", line 288, in parse
    iterator, tag, data, config, pos)
  File "/home/cbanek/pyvo/pyvo/utils/xml/elements.py", line 202, in add_complexcontent
    element.parse(iterator, config)
  File "/home/cbanek/pyvo/pyvo/io/vosi/vodataservice.py", line 287, in parse
    super(ParamHTTP, self).parse(iterator, config)
  File "/home/cbanek/pyvo/pyvo/utils/xml/elements.py", line 288, in parse
    iterator, tag, data, config, pos)
  File "/home/cbanek/pyvo/pyvo/utils/xml/elements.py", line 230, in add_simplecontent
    getattr(self, attr_name).append(data)
  File "/home/cbanek/pyvo-env/lib/python3.6/site-packages/astropy/utils/collections.py", line 47, in append
    self._assert(x)
[...]

PR #114, opened 2019-04-02
Fix SecurityMethod parsing of capabilities.

cbanek:
There was a problem where if any SecurityMethod was mentioned,
there would be a parse error since the SecurityMethod class wasn't
being used.  Reorder the classes and add the correct attribute.


cbanek:
Fixes #113

cbanek:
Fix a simple whitespace issue (add an extra blank line),  I'm not sure why the Travis check is broken, but doesn't seem to have to do with this PR.

funbaker:
Looks fine, merging.

Thank you!

PS: Build failing has nothing to do with this PR, looks like some API changed or smth.
Because of that i cannot do a release yet, can you install it from master for now?

PR #115, opened 2019-04-04
Fix pycodestyle errors

cbanek:
This should fix all the whitespace errors currently emitted by
pycodestyle for this repo.

Fix setup.cfg to say pycodestyle since pep8 is deprecated.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -8,7 +8,7 @@
 # Note that not all possible configuration values are present in this file.
 #
 # All configuration values have a default. Some values are defined in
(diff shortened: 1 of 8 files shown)


cbanek:
I know these can be annoying PRs, but I just went through and fixed the whitespace style errors from pycodestyle and updated it.  This should be a whitespace only change that should hopefully get a clean linter run.

cbanek:
Linter test now green.  Trying to understand the build break, but I'm pretty lost.

PR #116, opened 2019-04-04
Fix mimetype confusion during tests

cbanek:
It seems like it's easy for python to get confused when you name
a file the same as a dependent module.  This happened for mimetypes,
and seems to call the wrong mimetypes when trying to add_type the
second time.


cbanek:
I think if you also take PR #112 we can get back to passing?

funbaker:
Merged this and the referenced one. Lets see if this unbreaks the tests.

Thank you!

cbanek:
Well looks like we're closer to unbreaking the tests.  The other PR will get the style ones.  I'm still vexed as to why this didn't fix that last failing test.  It didn't fail in PR #112 's travis branch run!

funbaker:
I think it is related to https://github.com/astropy/ccdproc/pull/650/files

PR #118, opened 2019-04-23
MNT: Remove conftest.py

pllim:
Does not appear to have anything useful in the file and its existence is causing failure in astropy dev job.


pllim:
Nope. Trying something else...

PR #132, opened 2019-05-29
MNT: Use sphinx_astropy on conf.py

pllim:
This should get rid of import error when trying to build doc. The command to build doc should be:

```
python setup.py build_docs
```

For the record, `sphinx-astropy` is at https://github.com/astropy/sphinx-astropy


cbanek:
Fantastic thanks, just tested it out and it works.  Taking this.

PR #133, opened 2019-05-29
MNT: Update org in setup.cfg

pllim:
Update the org in `setup.cfg`.


cbanek:
Looks good, thanks!  Naming is always so dangerous, it gets everywhere!

PR #142, opened 2019-06-04
Remove unused code from utils/xml/exceptions.py

### LINKED ISSUE #141: MNT: Remove duplicate code in pyvo.utils.xml.exceptions
As brought up by @funbaker in https://github.com/astropy/pyvo/issues/130#issuecomment-498203485

*For later consideration: There are also those replicated methods here:*

https://github.com/astropy/pyvo/blob/3cb3434abc9ffb6559a323edef6e4d0783cca38a/pyvo/utils/xml/exceptions.py#L25

*(For some reason i created them but didnt used them in the xml parser)*

cbanek:
We used to be calling this code but now we just call the
astropy version instead.  So this copy is unused, so
we're deleting it and anything unused beneath it.

Fixes #141.

### DIFF
--- a/pyvo/utils/xml/exceptions.py
+++ b/pyvo/utils/xml/exceptions.py
@@ -3,13 +3,9 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
-from warnings import warn
-
 from astropy.utils.exceptions import AstropyWarning
 
 __all__ = ['XMLWarning', 'UnknownElementWarning']
-
-MAX_WARNINGS = 10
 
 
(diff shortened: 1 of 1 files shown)


cbanek:
Local tests pass, hoping we still get a green check mark on this one, then let the review begin!

cbanek:
Okay looks like this passes the tests, although I'd like a careful review by @funbaker just because with python and no static typing it can be hard to tell if these are used somewhere else.  No rush on this one, it's just a good cleanup.

cbanek:
OK thanks everyone, I'm going to take this!

PR #157, opened 2019-06-28
Auth

cbanek:
Sorry for making a different branch and review, but I am trying to keep things clean and have the ability to look at what I've done in the other branch.

I've changed a lot of things around based on feedback:

- Session is now the thing that is associated with a securitymethod URI, since in requests you can really do anything with a session.
- Session is passed into each object via keyword arg, and creates one if it doesn't exist.
- AuthSession now ducktypes for pyvo as a session, since pyvo only calls get and post.  By not having it subclass session, properties that could alter behavior have no effect.
- Helper methods for easy session configuration for cookies, basic auth, and certificates were added.
- AuthSession is passed in to attach it to a service.  It does this by looking for a special parse_capabilities attribute.
- I've moved the 'tests', which were really more examples, to the examples directory.  

If people think this is looking close, I will start writing unit tests against it that don't require a server to keep it in line with the other pyvo tests.  Also, updating doc strings, etc.

### DIFF
--- a/examples/auth/cadc_auth_example.py
+++ b/examples/auth/cadc_auth_example.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Example for authenticating with CADC TAP service
+"""
+import getpass
+
+import requests
+
+import pyvo
+import pyvo.extensions.auth.authsession
+
+# Gather login information
+data = {
+    'username': input('Username:'),
+    'password': getpass.getpass('Password:')
+}
+
+headers = {
+    'Content-Type': 'application/x-www-form-urlencoded',
+    'Accept': 'text/plain'
+}
+
+# Do the login and get the cookie
+login_url = 'https://ws-cadc.canfar.net/ac/login'
+response = requests.post(login_url, data=data, headers=headers)
+response.raise_for_status()
+cookie = '\"' + response.text + '\"'
+
+# Configure the session and run the query
+auth = pyvo.extensions.auth.authsession.AuthSession()
+auth.credentials.set_cookie('CADC_SSO', cookie)
+service = pyvo.dal.TAPService('https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/tap', auth)
+job = service.search('SELECT * from TAP_SCHEMA.tables')
+print(job)
--- a/examples/auth/cert_auth_example.py
+++ b/examples/auth/cert_auth_example.py
@@ -1,0 +1,15 @@
+#!/usr/bin/env python
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Example for using client certificate authentication with CADC TAP service
+Tests for pyvo.extensions.authsession
+"""
+import pyvo
+import pyvo.extensions.auth.authsession
+
+certificate_path = input('Path to client certificate file:')
+auth = pyvo.extensions.auth.authsession.AuthSession()
+auth.credentials.set_client_certificate(certificate_path)
+service = pyvo.dal.TAPService('https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/tap', auth)
+job = service.search('SELECT * from TAP_SCHEMA.tables')
+print(job)
--- a/examples/auth/gaia_auth_example.py
+++ b/examples/auth/gaia_auth_example.py
@@ -9,6 +9,7 @@ import requests
 
 import pyvo
 import pyvo.extensions.auth.authsession
+from pyvo.extensions.auth.securitymethods import ANONYMOUS
 
 # Gather login information
 data = {
@@ -30,7 +31,7 @@ response.raise_for_status()
 
 # Use this session with the auth cookie for all requests to Gaia.
 auth = pyvo.extensions.auth.authsession.AuthSession()
-auth.credentials.set('anonymous', session)
+auth.credentials.set(ANONYMOUS, session)
 service = pyvo.dal.TAPService('http://gea.esac.esa.int/tap-server/tap', auth)
 job = service.search('SELECT * from TAP_SCHEMA.tables')
 print(job)
--- a/examples/auth/gaia_auth_example.py
+++ b/examples/auth/gaia_auth_example.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Example for authenticating with Gaia TAP service
+"""
+import getpass
+
+import requests
+
+import pyvo
+import pyvo.extensions.auth.authsession
+
+# Gather login information
+data = {
+    'username': input('Username:'),
+    'password': getpass.getpass('Password:')
+}
+
+headers = {
+    'Content-Type': 'application/x-www-form-urlencoded',
+    'Accept': 'text/plain'
+}
+
+# Create a session and do the login.
+# The cookie will end up in the cookie jar of the session.
+login_url = 'http://gea.esac.esa.int/tap-server/login'
+session = requests.Session()
+response = session.post(login_url, data=data, headers=headers)
+response.raise_for_status()
+
+# Use this session with the auth cookie for all requests to Gaia.
+auth = pyvo.extensions.auth.authsession.AuthSession()
+auth.credentials.set('anonymous', session)
+service = pyvo.dal.TAPService('http://gea.esac.esa.int/tap-server/tap', auth)
+job = service.search('SELECT * from TAP_SCHEMA.tables')
+print(job)
--- a/pyvo/dal/tap.py
+++ b/pyvo/dal/tap.py
@@ -21,7 +21,7 @@ from ..io import vosi, uws
 from ..io.vosi import tapregext as tr
 
 from ..utils.formatting import para_format_desc
-from ..utils.http import session
+from ..utils.http import use_session
 
 __all__ = [
     "search", "escape", "TAPService", "TAPQuery", "AsyncTAPJob", "TAPResults"]
@@ -96,7 +96,7 @@ class TAPService(DALService, AvailabilityMixin, CapabilityMixin):
         if self._tables is None:
             tables_url = '{}/tables'.format(self.baseurl)
 
-            response = session.get(tables_url, stream=True)
+            response = self._session.get(tables_url, stream=True)
 
             try:
                 response.raise_for_status()
@@ -230,7 +230,7 @@ class TAPService(DALService, AvailabilityMixin, CapabilityMixin):
         AsyncTAPJob
         """
         job = AsyncTAPJob.create(
-            self.baseurl, query, language, maxrec, uploads, **keywords)
+            self.baseurl, query, language, maxrec, uploads, self._session, **keywords)
         job = job.run().wait()
         job.raise_if_error()
         result = job.fetch_result()
@@ -267,7 +267,7 @@ class TAPService(DALService, AvailabilityMixin, CapabilityMixin):
         AsyncTAPJob
         """
         return AsyncTAPJob.create(
-            self.baseurl, query, language, maxrec, uploads, **keywords)
+            self.baseurl, query, language, maxrec, uploads, self._session, **keywords)
 
     def create_query(
             self, query=None, mode="sync", language="ADQL", maxrec=None,
@@ -295,7 +295,7 @@ class TAPService(DALService, AvailabilityMixin, CapabilityMixin):
             a mapping from table names to objects containing a votable.
         """
         return TAPQuery(
-            self.baseurl, query, mode, language, maxrec, uploads, **keywords)
+            self.baseurl, query, mode, language, maxrec, uploads, self._session, **keywords)
 
     def describe(self, width=None):
         """
@@ -335,7 +335,7 @@ class AsyncTAPJob:
     @classmethod
     def create(
             cls, baseurl, query, language="ADQL", maxrec=None, uploads=None,
-            **keywords):
+            session=None, **keywords):
         """
         creates a async tap job on the server under `baseurl`
 
@@ -352,15 +352,17 @@ class AsyncTAPJob:
             the maximum records to return. defaults to the service default
         uploads : dict
             a mapping from table names to objects containing a votable
+        session : object
+           optional session to use for network requests
         """
         query = TAPQuery(
             baseurl, query, mode="async", language=language, maxrec=maxrec,
-            uploads=uploads, **keywords)
+            uploads=uploads, session=session, **keywords)
         response = query.submit()
-        job = cls(response.url)
+        job = cls(response.url, session=session)
         return job
 
-    def __init__(self, url):
+    def __init__(self, url, session=None):
         """
         initialize the job object with the given url and fetch remote values
 
@@ -370,6 +372,7 @@ class AsyncTAPJob:
             the job url
         """
         self._url = url
+        self._session = use_session(session)
         self._update()
 
     def __enter__(self):
@@ -393,13 +396,13 @@ class AsyncTAPJob:
         """
         try:
             if wait_for_statechange:
-                response = session.get(
+                response = self._session.get(
                     self.url, stream=True, timeout=timeout, params={
                         "WAIT": "-1"
                     }
                 )
             else:
-                response = session.get(self.url, stream=True, timeout=timeout)
+                response = self._session.get(self.url, stream=True, timeout=timeout)
             response.raise_for_status()
         except requests.RequestException as ex:
             raise DALServiceError.from_except(ex, self.url)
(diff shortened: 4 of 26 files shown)


examples/auth/cadc_auth_example.py
```diff
@@ -0,0 +1,35 @@
...
+login_url = 'https://ws-cadc.canfar.net/ac/login'
+response = requests.post(login_url, data=data, headers=headers)
+response.raise_for_status()
+cookie = '\"' + response.text + '\"'
+
+# Configure the session and run the query
+auth = pyvo.extensions.auth.authsession.AuthSession()
+auth.credentials.set_cookie('CADC_SSO', cookie)
+service = pyvo.dal.TAPService('https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/tap', auth)
+job = service.search('SELECT * from TAP_SCHEMA.tables')
```
andamian: For power users I would maybe add an optional `session` of type `requests.session` attribute to the `search` or other methods in the class that hit the server. This attribute would override the internal `session(s)` created by the object to allow possible optimizations. This is optional and we should add it only if there are clear scenarios that requires it.

examples/auth/gaia_auth_example.py
```diff
@@ -0,0 +1,36 @@
...
+}
+
+# Create a session and do the login.
+# The cookie will end up in the cookie jar of the session.
+login_url = 'http://gea.esac.esa.int/tap-server/login'
+session = requests.Session()
+response = session.post(login_url, data=data, headers=headers)
+response.raise_for_status()
+
+# Use this session with the auth cookie for all requests to Gaia.
```
andamian: It's relatively easy to use the cookies jar in this case:
```
service = pyvo.dal.TAPService('http://gea.esac.esa.int/tap-server/tap', auth={'SSO_COOKIE': session.cookies})
```
`session.cookies` is a `CookieJar` which should contain the newly acquired GAIA cookie. Disclaimer: I haven't tested it.

There might be other methods to pass in your own `CookieJar` in `requests.get` and have it augmented with new cookies without creating the session first but I couldn't make it work.

examples/auth/cadc_auth_example.py
```diff
@@ -0,0 +1,35 @@
...
+    'Content-Type': 'application/x-www-form-urlencoded',
+    'Accept': 'text/plain'
+}
+
+# Do the login and get the cookie
+login_url = 'https://ws-cadc.canfar.net/ac/login'
+response = requests.post(login_url, data=data, headers=headers)
+response.raise_for_status()
+cookie = '\"' + response.text + '\"'
+
```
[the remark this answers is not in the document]
cbanek: Right, I agree.  I moved these into an examples folder instead, because I think people will want to see how these things work, and examples are good.  But the details of how to get the cookie, get the certificate, etc, are external to the framework of doing auth in pyvo.

examples/auth/cadc_auth_example.py
```diff
@@ -0,0 +1,35 @@
...
+# Configure the session and run the query
+auth = pyvo.extensions.auth.authsession.AuthSession()
+auth.credentials.set_cookie('CADC_SSO', cookie)
+service = pyvo.dal.TAPService('https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/tap', auth)
```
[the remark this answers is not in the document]
cbanek: Well I can make the import names shorter, or just put it in the global namespace if you want.  I think that's reasonable.

As I believe I mentioned on the previous PR and in our slack conversations, I very much disagree with the concept of passing a dictionary in for this as opposed to an object, which I will detail in a much longer later comment.

examples/auth/cert_auth_example.py
```diff
@@ -0,0 +1,17 @@
...
+
+import pyvo
+import pyvo.extensions.auth.authsession
+
+certificate_path = input('Path to client certificate file:')
+auth = pyvo.extensions.auth.authsession.AuthSession()
+auth.credentials.set_client_certificate(certificate_path)
+service = pyvo.dal.TAPService('https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/tap', auth)
+job = service.search('SELECT * from TAP_SCHEMA.tables')
+print(job)
```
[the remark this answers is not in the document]
cbanek: > auth can have a combination of multiple authorization methods and the class will try to match them with the ones supported by the service and create the appropriate internal sessions.
Right, that's what it does right now with that object, it matches the authorization methods against the ones supported by the service and uses the right session.

BasicAA is already supported by using the set_password function.  It uses the requests.sessions auth setting the tuple to username password.

> For extensions we could follow the requests model:
Yes, the idea is to allow people to put in their own sessions, with whatever authbase they want.  The problem I discovered in the previous version of this PR is that trying to have multiple authbases on one session is hard.  And some things like certificate auth are outside of the authbase part.  So making the session available is the safest bet to allow people to do whatever they want, including using whatever requests authenticators might be out there.

I'm not sure if these comments are calling for something, because this is already how it is in the code up for review?

examples/auth/cadc_auth_example.py
cbanek: The classes themselves already take the session parameter, but you're right that the some of the highest level wrappers that try to wrap everything into one line don't have this.  But someone could just make the class themselves - it's not like we're preventing something, this is just about lines of code.  I could add in the parameter, but I'm not sure anyone is calling for it.

examples/auth/gaia_auth_example.py
cbanek: Right, that's what I'm doing here right now.  I'm passing in the session, and its cookie jar, which has the cookie.

I also have the set_cookie function in credstore.  I could make a function that takes a cookie jar, that is also good.

cbanek:
Okay so here I will discuss my reasoning to use an object over a dict, and why I think this is the right design decision:

First off, auth is complicated.  It's error prone, it's complicated, it involves secrets and possibly exposing secrets.  So we have to be careful.  Things need to be very obvious and explicit.  With this in mind...

A dict has some nice properties.  It's an easy to work with data type, and it can contain pretty much anything.  But in the end, it is a data type, and holds data - not logic.  If I were passing in a dict as you say, I would still need all the layers underneath it to actually do the logic.  So this doesn't really change the underlying implementation, but only the way it looks to the caller.

(Note: Again, just the creds isn't enough, we also need the capabilities of the service, and all this data needs to be passed down to child objects like TAPResults, etc. since they don't have access to the capabilities of the service who created them otherwise)

Now, is allowing the user to pass in a dict a good user interface?  I would argue that an object is a better interface.  In the object I have created, I have functions that people can look up (eventually with docstrings) that will say what functions do what, and the functions are very simply named.  This would be opposed to the schema of the dict, which could be complicated, and harder to document.  Given that python is very type loose, understanding this schema is everything.  For example, in this PR you said that we could pass in SSO_COOKIE both as a cookiejar and a dict.  I think it's less cognitive load to just call a function and pass in either the cookiejar or the cookie, and I can validate that what is passed in is the expected type.  If not, we could throw an exception right there.

The data model of this schema is also likely to be complex and need to be changed, at the very least to add new keys to support new auth methods, and plumbing that down through.  With the PR currently as posted, someone could write a new auth method and attach it to an auth URI and have it work with no changes to pyvo.  I think that's great.

This may be a bit of my PTSD from Kubernetes where it takes yaml, and you need to discover all the keys, and it will sometimes silently do bad things for you, but I find an actual interface with functions and objects to be much preferable to trying to shove everything into a data model where what each key does is less clear.  Many times to understand what I can pass in to some of these dicts or kwargs, I need to look deeply into the code.  I really don't want that to be the case with auth.

examples/auth/cadc_auth_example.py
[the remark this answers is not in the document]
andamian: I don't have a strong opinion on the dictionary except that in Python it seems to be the defacto way of passing data structures. The `CredentialStore` class has its own advantages as you've list them so I'm fine with that (in `opencadc` we also use a class to pass user credentials). I would still however build the `AuthSession` inside the `TAPService` constructor unless there is a good reason to leave it on the user to do that. The code could become something like:
```
from pyvo.extensions import CredentialStore
cred = CredentialStore()
cred.set_cookie(('name', 'value'))  # or cred.set_cookie(cookie_jar)
cred.set_client_certificate('certificate_path')
cred.set_auth_session('key', session)
tap = pyvo.dal.TAPService('https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/tap', auth)
```

examples/auth/gaia_auth_example.py
andamian: `set_cookie` (and optionally adding `set_cookie_jar`) in `CredStore` that could handle cookie values and cookie jars is more consistent and therefore preferable.

examples/auth/cert_auth_example.py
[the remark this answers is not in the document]
andamian: It was just an example to show that the API allows for other cases to be added later on.

examples/auth/cadc_auth_example.py
andamian: The classes do take the session parameter but is that session meant to be used with all the end points of the service? Passing it with the actual function allows the caller to target specific end point(s).

examples/auth/gaia_auth_example.py
cbanek: Okay I'll add a set_cookie_jar as a function to the credstore

examples/auth/cadc_auth_example.py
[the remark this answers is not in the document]
cbanek: So the only thing I see as different from what your code is and what I'm doing in the example is that it gets rid of calling into a member of the subclass of authsession.  If I flatten this out (put all the credstore logic into authsession, and then call it credstore) would that be fine?  I think that's the only difference I can see.

The reason I don't want to compose them inside of the TAP service is that then I would need to compose it the same way in every subobject.  For example, TAPService makes objects like AsyncTAPResults and TAPResults which use the sessions.  What they need is the creds, and the mapping (can only be gotten from tap service).  So TAPService needs to pass all this info down.  It currently passes it as one object, the authsession.

Composing the object everywhere it is used is somewhat trouble, since I don't have all the info required to make it.  So every child helper class would need to take an authsession.  In this way, at least it's consistent that every class takes a session.

In the way I've done it, there are theoretically very limited changes for people who aren't using auth (they are just not calling a global session anymore).  Then we simply pass in a class that also works as a session.  Someone could even pass in something other than authsession, like a different requests.Session entirely if they wanted to.  In this way, the classes like TAPService and all its child classes don't need to do anything to session, they just call it (either GET or POST, that's it).

examples/auth/cadc_auth_example.py
cbanek: Sorry what is an endpoint in this context?  Auth will only be used on URLs that call for it in the capabilities.  Otherwise you have to set the anonymous session, which will be used for all the other calls (it can still do auth, or whatever it needs to, but it is used when the negotiated security method is anonymous).

cbanek:
Okay, lots of changes:
1. Tests
2. Docstrings updated
3. Add cookie_jar setting in credstore
4. Fix test problem with loading pkg resources.

Build is green.

pyvo/dal/tap.py
```diff
@@ -77,16 +77,25 @@ class TAPService(DALService, AvailabilityMixin, CapabilityMixin):
...
         """
-        super().__init__(baseurl)
+        super().__init__(baseurl, session=session)
+
+        # Check if the session has a parse_capabilities attribute.
+        # This means that the session is aware of IVOA capabilities,
+        # and can use this information in processing network requests.
+        # One such usecase for this is auth.
+        if hasattr(self._session, 'parse_capabilities'):
+            self._session.parse_capabilities(self.capabilities)
```
funbaker: Not sure about this one, there is already a method named `parse_capabilities` in the xml code.

cbanek: Well nobody has ever said I'm overly creative about variable names. :smile:
I've gone ahead and changed this to update_from_capabilities

cbanek:
Okay, some new changes.

1. Change parse_capabilities to update_from_capabilities (just a naming change)
2. I found some issues the other day with how hitting the capabilities and availability endpoints from the mixins weren't using the sessions and were using requests directly.  So I made these calls use the same session variable on the class. (also will help with things like user-agent)
3. I've also added a function to authsession to wrap authurls, where someone can add a URL and associate it with a security method outside of capabilities.  While the capabilities is nice to do the negotiation, sometimes you want to set it for services that don't advertise, or don't have capabilities, but you know you want to use a particular auth method.  This provides a good catch all to force negotiation.

pyvo/dal/tap.py
funbaker: Was just something which confused me at some point.

timj:
I do prefer a class to a dict since it does give you more control.

@andamian it looks to me like @cbanek has dealt with your review comments. Can you please respond and indicate whether you approve the current version or feel there is a fundamental problem with the PR. We would like to get unblocked on this but the PR requires at least one approval from someone.

 @bmerin or @jcsegovia do you have any intention of reviewing?  Auth is really important for the integration of ESAC into pyvo so it would be great if you could sign off.

andamian:
> I do prefer a class to a dict since it does give you more control.
> 
> @andamian it looks to me like @cbanek has dealt with your review comments. Can you please respond and indicate whether you approve the current version or feel there is a fundamental problem with the PR. We would like to get unblocked on this but the PR requires at least one approval from someone.
> 
> @bmerin or @jcsegovia do you have any intention of reviewing? Auth is really important for the integration of ESAC into pyvo so it would be great if you could sign off.

Sorry - I was busy/away in vacation. I'm in the process of reviewing the latest changes. At one point I think @msdemlei also expressed an interest in having a look. I agree that we should get this closed - I'd like to start using it before the release of 1.0.

cbanek:
Okay pushed a new commit to get out some of the string constants for URIs for different security methods.  Hopefully we can now call this done.

cbanek:
Excellent, thank you, merging this.

msdemlei:
Apologies from me, too, for having procrastinated my review of this
PR.

So, here's twp belated thoughts:

(a) I'd not use accessURL/@use to determine whether or not the auth
should be sent to child resources; the semantics of @use are murky at
best.  I *think* most people understood that use="base" meant you
were supposed to append urlencoded arguments and then run GET,
whereas use="full" meant you either aren't supposed to pass arguments
or perhaps POST them.  I doubt anyone is doing anything useful with
it, which suggests it's going to be weird.  

Don't take my word for it -- run

```
select distinct standard_id
from rr.interface natural join rr.capability
where url_use='base'
```

on a RegTAP server, and then with url_use='full'.  If we're sure we
need something like "this service will respond to child resources of
the access url, too" and want to use @use for that, we'll have to
clean up existing usage, too.  Possible, but not something I look for
given my recent experience with a much smaller problem.

Given we're mostly dealing with TAP or other services that can have
async (i.e., UWS) interfaces here, I think anything but sending
credentials to all child resources as well is bound to be a huge
pain.  I'd suggest to forget about the whole full_urls/base_urls
distinction; but if you make it, don't depend on accessURL/@use.

You may still want the logic to select the longest-match URL, though
I'd say nested resources with differing authentication requirements
should really be redesigned.

(b) I had promised to think about ways to get around per-service
auth classes, but these aren't part of this PR.  Can anyone point me
at them?

(c) By VOResource 1.1, there's now mirror URLs, too; we'll have to
think about how that affects this at some point.  I guess it should
be ok to expose them in voresource.Capabilities and then iterate over
itertools.chain(i.accessurls, i.mirrorurls).  But IMHO that can wait
until there is a service that has both authentication and mirrors,
which I don't see as imminent; it would suck if we add logic here and
then, say, the mirrors end up requiring different credentials...

      -- Markus

funbaker:
> (c) By VOResource 1.1, there's now mirror URLs, too; we'll have to think about how that affects this at some point. I guess it should be ok to expose them in voresource.Capabilities and then iterate over itertools.chain(i.accessurls, i.mirrorurls). But IMHO that can wait until there is a service that has both authentication and mirrors, which I don't see as imminent; it would suck if we add logic here and then, say, the mirrors end up requiring different credentials...

This explains those UnknownElementWarning in https://github.com/astropy/pyvo/issues/165
(So `mirrorURL` needs to be implemented first anyway)

esdc-esac-esa-int:
Both anonymous and authenticated accesses to the Gaia Archive through pyvo.tap work fine. As there were changes between implementations of pyvo.tap and astroquery.utils.tap modules, we are still going through the assessment of how this change impacts the astroquery.gaia module.

andamian:
> Both anonymous and authenticated accesses to the Gaia Archive through pyvo.tap work fine. As there were changes between implementations of pyvo.tap and astroquery.utils.tap modules, we are still going through the assessment of how this change impacts the astroquery.gaia module.

We've already converted `astroquery.cadc` to use `pyvo` if you want to have an example. `pyvo` is still an optional package in `astroquery` but with `pyvo` 1.0 release and more usage in the subpackages we can argue for it to become a dependency.

PR #164, opened 2019-08-23
Fix AstropyDeprecationWarning in SIA & SSA

### LINKED ISSUE #163: AstropyDeprecationWarning in SSAService & SIAService
Hi all, 
I guess for now this is more of a heads up and I hope here is the right place to bring it up:
I encounter a AstropyDeprecationWarning when I use `pyvo.dal.SIAService` and the `pyvo.dal.SSAService`. Here is an example of the code:

`from astropy.coordinates import SkyCoord`
`from astropy import units as u`
`import pyvo`
`ssa_service = pyvo.dal.SSAService('https://archive.stsci.edu/ssap/search2.php?id=HST&')`
`position = SkyCoord.from_name('A1656')`
`diameter = u.Quantity(2 * 40.0, unit="arcmin")`
`results = ssa_service.search(pos=position, diameter=diameter)`

The warning I get is the following (plus a bunch of VO Table warnings because Astropy is not happy with the units in the table):

`WARNING: AstropyDeprecationWarning: The truth value of a Quantity is ambiguous. In the future this will raise a ValueError. [astropy.units.quantity]`

The same warning comes up for the services in the documentation for both SSAService and SIAService. This might have to do with the Astropy issues discussed here https://github.com/astropy/astropy/issues/6580 and here https://github.com/astropy/astropy/pull/6590.

cbanek:
Fixes #163

When running SSAService.search() and passing in an astropy Quantity,
this warning would be emitted:

WARNING: AstropyDeprecationWarning: The truth value of a Quantity is ambiguous.
In the future this will raise a ValueError. [astropy.units.quantity]

Instead of checking the truth value of the parameter, I'm just going to check
if it is None or not, not relying on what the truth value of the Quantity
may be.

### DIFF
--- a/pyvo/dal/sia.py
+++ b/pyvo/dal/sia.py
@@ -399,7 +399,7 @@ class SIAQuery(DALQuery):
         if pos:
             self.pos = pos
 
-        if size:
+        if size is not None:
             self.size = size
 
         if format:
--- a/pyvo/dal/ssa.py
+++ b/pyvo/dal/ssa.py
@@ -349,7 +349,7 @@ class SSAQuery(DALQuery):
         if pos:
             self.pos = pos
 
-        if diameter:
+        if diameter is not None:
             self.diameter = diameter
 
         if band:


cbanek:
Seems simple enough for a no changelog bug since we're just fixing an emitted warning.

jazamatazz:
I'm getting an AstropyDeprecationWarning again.

PR #169, opened 2019-08-28
Implementation of get_job_list

### LINKED ISSUE #150: List jobs functionality missing in TAP
Users need a way to list their own TAP jobs. This will only work when auth is supported (#122)

andamian:
Resolve #150

### DIFF
--- a/pyvo/dal/tap.py
+++ b/pyvo/dal/tap.py
@@ -26,6 +26,24 @@ from ..utils.http import use_session
 __all__ = [
     "search", "escape", "TAPService", "TAPQuery", "AsyncTAPJob", "TAPResults"]
 
+IVOA_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+
+
+def _from_ivoa_format(datetime_str):
+    """
+    parses an ivoa date in ISO 8601 format: YYYY-MM-DDTHH:MM:SS.[mmm]Z
+
+    :param datetime_str:
+    :return: corresponding datetime object
+    """
+    # TODO Replace with datetime.fromisoformat(date_string) in Python3.7+
+    try:
+        # with fraction of seconds first
+        return datetime.strptime(datetime_str, IVOA_DATETIME_FORMAT)
+    except ValueError:
+        # and without
+        return datetime.strptime(datetime_str, "%Y-%m-%dT%H:%M:%SZ")
+
 
 def escape(term):
     """
@@ -306,6 +324,75 @@ class TAPService(DALService, AvailabilityMixin, CapabilityMixin):
         return TAPQuery(
             self.baseurl, query, mode, language, maxrec, uploads, self._session, **keywords)
 
+    def get_job(self, job_id):
+        """
+        Returns the job corresponding to an ID. Note that the caller must be
+        able to see the job in the current security context.
+
+        Parameters
+        ----------
+        job_id : str
+            ID of the job to view
+
+        Returns
+        -------
+        `~pyvo.io.vosi.endpoint.JobSummary` corresponding to the job ID
+        """
+        response = self._session.get(
+            self.baseurl + '/async/' + job_id,
+            stream=True)
+        response.raw.read = partial(response.raw.read,
+                                    decode_content=True)
+        return uws.parse_job(response.raw.read)
+
+    def get_job_list(self, phases=None, after=None, last=None,
+                     short_description=True):
+        """
+        lists jobs that the caller can see in the current security context.
+        The list can be filtered on the server side by the phases of the jobs,
+        creation date time or
+        Note that by default jobs in 'ARCHIVED` phase are not returned.
+
+        Parameters
+        ----------
+        phases: list of str
+            Union of job phases to filter the results by.
+        after: datetime
+            Return only jobs created after this datetime
+        last: int
+            Return only the most recent number of jobs
+        short_description: flag - True or False
+            If True, the jobs in the list will contain only the information
+            corresponding to the TAP ShortJobDescription object (job ID, phase,
+            run ID, owner ID and creation ID) whereas if False, a separate GET
+            call to each job is performed for the complete job description.
+
+        Returns
+        -------
+        list of `~pyvo.io.vosi.endpoint.JobSummary`
+        """
+
+        params = {'PHASE': phases, 'LAST': last}
+
+        if after:
+            if isinstance(after, str):
+                after = _from_ivoa_format(after)
+            params['AFTER'] = after.strftime(IVOA_DATETIME_FORMAT)
+
+        response = self._session.get('{}/async'.format(self.baseurl),
+                                     params=params,
+                                     stream=True)
+        response.raw.read = partial(response.raw.read, decode_content=True)
+
+        jobs = uws.parse_job_list(response.raw.read)
+        if not short_description:
+            dj = []
+            for job in jobs:
+                dj.append(self.get_job(job.jobid))
+            return dj
+        else:
+            return list(jobs)
+
     def describe(self, width=None):
         """
         Print a summary description of this service.
@@ -492,16 +579,13 @@ class AsyncTAPJob:
         value : datetime
             datetime after which the job results are deleted automatically
         """
-        try:
-            # is string? easier to ask for forgiveness
-            value = datetime.strptime(value, "%Y-%m-%dT%H:%M:%SZ")
-        except ValueError:
-            pass
+        if isinstance(value, str):
+            value = _from_ivoa_format(value)
 
         try:
             response = self._session.post(
                 "{}/destruction".format(self.url),
-                data={"DESTRUCTION": value.strftime("%Y-%m-%dT%H:%M:%SZ")})
+                data={"DESTRUCTION": value.strftime(IVOA_DATETIME_FORMAT)})
             response.raise_for_status()
         except requests.RequestException as ex:
             raise DALServiceError.from_except(ex, self.url)
--- a/pyvo/dal/tests/test_tap.py
+++ b/pyvo/dal/tests/test_tap.py
@@ -4,6 +4,7 @@ Tests for pyvo.dal.tap
 """
 from functools import partial
 from contextlib import ExitStack
+import datetime
 import re
 from io import BytesIO
 from urllib.parse import parse_qsl
@@ -75,11 +76,21 @@ class MockAsyncTAPServer:
                 )),
                 'result': stack.enter_context(mocker.register_uri(
                     'GET', job_re_result_full, content=self.result
+                )),
+                'get_job': stack.enter_context(mocker.register_uri(
+                    'GET', 'http://example.com/tap/async/111',
+                    content=self.get_job
+                )),
+                'get_job_list': stack.enter_context(mocker.register_uri(
+                    'GET', 'http://example.com/tap/async',
+                    content=self.get_job_list
                 ))
             }
             yield matchers
 
     def create(self, request, context):
+        if request.method == 'GET':
+            return self.get_job_list(request, context)
         self.validator(request)
         newid = max(list(self._jobs.keys()) or [0]) + 1
         data = dict(parse_qsl(request.body))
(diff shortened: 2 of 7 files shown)


pyvo/io/uws/endpoint.py
```diff
@@ -9,22 +9,27 @@
...
+from .tree import JobSummary, JobList
 
 __all__ = ["parse_job", "JobFile"]
 
 
-def parse_job(
-    source, pedantic=None, filename=None, _debug_python_based_parser=False
+
+
+def parse_for_object(
```
funbaker: It's a good idea to avoid redundant functions, but it makes the code (or the library) less readable.
Maybe create two functions with `functools.partial`?

Edit: Maybe even put the generic function into `utils.xml`.

pyvo/io/uws/tree.py
```diff
@@ -260,6 +260,28 @@ def results(self, iterator, tag, data, config, pos):
...
         self._results = results
 
+class JobListElem(JobSummary):
+
+    def __init__(self, config=None, pos=None, _name='parameters', **kwargs):
+        JobSummary.__init__(self, config, pos, _name, **kwargs)
+        self._jobid = kwargs.get('id')
+
+
+class JobList(JobListElem, HomogeneousList):
```
funbaker: Is it intended to give `JobList` the same attributes etc. like `JobListElem`?

cbanek:
Probably want to undo the merged with master commit and rebase instead and then force push

andamian:
> Probably want to undo the merged with master commit and rebase instead and then force push

Of course - habits. Thanks for pointing out...

pyvo/dal/tap.py
```diff
@@ -306,6 +309,84 @@ def create_query(
...
+                after = datetime.strptime(after, IVOA_TIME_FORMAT)
+            query += 'AFTER={}&'.format(after.strftime(IVOA_TIME_FORMAT))
+
+        if last:
+            query += 'LAST={}'.format(last)
+
+        if query:
+            query = '?{}'.format(query.strip('&'))
+
+        response = self._session.get('{}/async{}'.format(self.baseurl, query),
```
cbanek: Requests will let you pass query params like this as a dict, which means you hopefully won't have to worry about this kind of string parsing and query building:

https://2.python-requests.org//en/master/user/quickstart/#passing-parameters-in-urls

pyvo/dal/tap.py
```diff
@@ -501,6 +582,7 @@ def destruction(self, value):
         try:
             response = self._session.post(
                 "{}/destruction".format(self.url),
+                # wondering if this is not IVOA_DATETIME_FORMAT - microsec?
```
cbanek: Yeah seems close - should we try to change this to use your constant?  It does have more precision.  I'm guessing this is listed somewhere in the spec as the format of the time?

pyvo/dal/tests/test_tap.py
```diff
@@ -190,6 +201,67 @@ def result(self, request, context):
...
+        jobid = int(job_re_path.match(request.path).group(1))
+
+        job = JobFile()
+        job.jobid = jobid
+        job.phase = 'EXECUTING'
+        job.ownerid = '222'
+        job.creationtime = Time.now()
+        io = BytesIO()
+        job.to_xml(io)
+        return io.getvalue()
```
cbanek: I dig this pattern of being able to make results without hardcoded files.

pyvo/dal/tests/test_tap.py
```diff
@@ -190,6 +201,67 @@ def result(self, request, context):
...
+
+    def _get_jobref_rep(self, jobid, phase, runid, ownerid, creation_time):
+        doc = '    <uws:jobref id="{}">\n'
+        doc += '        <uws:phase>{}</uws:phase>\n'
+        doc += '        <uws:runId>{}</uws:runId>\n'
+        doc += '        <uws:ownerId>{}</uws:ownerId>\n'
+        doc += '        <uws:creationTime>{}</uws:creationTime>\n'
+        doc += '    </uws:jobref>\n'
```
cbanek: Maybe collapse this down to a multi line string rather than doing this appending?

andamian:
@funbaker and @cbanek - the PR is ready for review - please have a look. One thing that I'm not 100% sure, is the returned types in the job lists. Didn't know whether to create a separate class for this. For now, it returns a list of `JobSummary` objects. The level of information in these objects depends on the `short_description` flag (more attributes are filled in when the flag is set to `False`). I've also considered returning 'AsyncTAPJob' objects (which I would have preferred), except that they seem to access the server frequently. Opinions?

pyvo/utils/xml/elements.py
```diff
@@ -3,13 +3,65 @@
...
+
+    if filename is None and isinstance(source, str):
+        config['filename'] = source
+
+    with iterparser.get_xml_iterator(
+            source,
+            _debug_python_based_parser=_debug_python_based_parser
+    ) as iterator:
+        return object_type(
+            config=config, pos=(1, 1)).parse(iterator, config)
```
cbanek: Thanks for not duplicating this one :)

pyvo/utils/xml/elements.py
```diff
@@ -125,7 +177,7 @@ def object_attrs(obj):
     objtype = type(obj)
     attrs = {
         getattr(objtype, name).name: value for name, value in getmembers(obj)
-        if isinstance(getattr(objtype, name, None), xmlattribute)}
+        if value and isinstance(getattr(objtype, name, None), xmlattribute)}
```
cbanek: What's the story with this and :194?  Seems like there was maybe some parsing trouble?

pyvo/io/uws/__init__.py
```diff
@@ -1,4 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-__all__ = ['parse_job', 'JobFile']
+__all__ = ['parse_job', 'parse_for_object', 'JobFile', 'JobList']
 
-from .endpoint import parse_job, JobFile
+from .endpoint import *
+from .tree import *
```
cbanek: It seems like your doc build is broken, I'm no sphinx expert at all, but it looks like it's complaining about uwstree, and it might be just walking everything that's imported.  But the error looks strange and I don't know what it really means.

pyvo/dal/tests/test_tap.py
andamian: Not sure I understand...

pyvo/dal/tests/test_tap.py
andamian: Done

pyvo/dal/tap.py
andamian: Doh! What was I thinking...

pyvo/dal/tests/test_tap.py
cbanek: I just like this way of reusing the same code to make the XML.  I think it looks better than having the strings of XML, although I can understand why you'd want to do that too.  So I'm fine with either.  I think in the end, this whole test_tap thing might end up being a tap server in and of itself :wink:

No action necessary!

cbanek:
I think AsyncTAPJob makes sense, but if it takes a lot of network requests, that can get really chatty if you're looking over the last 100 jobs or something.  So I think what you've done here is more than fine.  Maybe if you want you could add some method to convert a JobSummary into an AsyncTAPJob?  I'm guessing this is just plumbing the URL through.

pyvo/dal/tests/test_tap.py
andamian: Agree! I've just followed the pattern.

pyvo/io/uws/__init__.py
andamian: This is a bugger. I can't reproduce it locally... I'll keep an eye on this - thanks for the tip.

pyvo/dal/tests/test_tap.py
funbaker: > I just like this way of reusing the same code to make the XML. I think it looks better than having the strings of XML, although I can understand why you'd want to do that too. So I'm fine with either. I think in the end, this whole test_tap thing might end up being a tap server in and of itself wink
> 
> No action necessary!

It's a bit more complicated than that (ADQL Parsing), but it's a start...

pyvo/io/uws/endpoint.py
```diff
@@ -4,14 +4,49 @@
...
 from astropy.utils.xml.writer import XMLWriter
 from astropy.io.votable.util import convert_to_writable_filelike
 
-from ...utils.xml.elements import xmlattribute
-from .tree import JobSummary
+from ...utils.xml.elements import xmlattribute, parse_for_object
+from .tree import JobSummary, JobList
 
-__all__ = ["parse_job", "JobFile"]
+__all__ = ["parse_job", "parse_for_object", "JobFile", "JobList"]
```
funbaker: Exposing a imported object.

pyvo/dal/tap.py
```diff
@@ -306,6 +309,76 @@ def create_query(
...
+        Returns
+        -------
+        `~pyvo.io.vosi.endpoint.JobSummary` corresponding to the job ID
+        """
+        response = self._session.get(
+            self.baseurl + '/async/' + job_id,
+            stream=True)
+        response.raw.read = partial(response.raw.read,
+                                    decode_content=True)
+        return uws.parse_for_object(response.raw.read, uws.JobSummary)
```
funbaker: Use `parse_for_object` from `utils`?

pyvo/dal/tap.py
```diff
@@ -306,6 +309,76 @@ def create_query(
...
+            if isinstance(after, str):
+                after = datetime.strptime(after, IVOA_TIME_FORMAT)
+            params['AFTER'] = after.strftime(IVOA_TIME_FORMAT)
+
+        response = self._session.get('{}/async'.format(self.baseurl),
+                                     params=params,
+                                     stream=True)
+        response.raw.read = partial(response.raw.read, decode_content=True)
+
+        jobs = uws.parse_for_object(response.raw.read,
```
funbaker: Same.

pyvo/io/uws/tree.py
```diff
@@ -261,7 +261,26 @@ def results(self, iterator, tag, data, config, pos):
         self._results = results
 
 
+class JobList(JobSummary, HomogeneousList):
```
funbaker: Does `JobList` got the same attributes as `JobSummary`?

funbaker:
> I think AsyncTAPJob makes sense, but if it takes a lot of network requests, that can get really chatty if you're looking over the last 100 jobs or something. So I think what you've done here is more than fine. Maybe if you want you could add some method to convert a JobSummary into an AsyncTAPJob? I'm guessing this is just plumbing the URL through.

`AsyncTAPJob` is for actual job control which depends on up2date data, while `JobSummary` is just a data structure. Conversion methods sounds fine.

pyvo/dal/tap.py
andamian: It turns out both formats (with and without fraction of seconds) are correct so I've added a parser to deal with this. This is temporary, as Python3.7+ already has a method to deal with iso datetime formats.

pyvo/dal/tap.py
andamian: Oops! Fixed.

pyvo/io/uws/endpoint.py
andamian: Fixed

pyvo/io/uws/tree.py
andamian: Good catch. Fixed this too. I went and fixed the `Element.parse` method to return the element - otherwise it was causing some grief.

andamian:
> > I think AsyncTAPJob makes sense, but if it takes a lot of network requests, that can get really chatty if you're looking over the last 100 jobs or something. So I think what you've done here is more than fine. Maybe if you want you could add some method to convert a JobSummary into an AsyncTAPJob? I'm guessing this is just plumbing the URL through.
> 
> `AsyncTAPJob` is for actual job control which depends on up2date data, while `JobSummary` is just a data structure. Conversion methods sounds fine.

I'm not sure what the value of the conversion code will be since as soon as the caller is trying to access some of the attributes of the `AsyncTAPJob` object will lead to network requests. If you can give me an example of a scenario where this would be useful, I will add it.

cbanek:
I think it's fine without it.  People can make their own lightweight conversion methods to do what they want if they need to.  Also great fix on that doc thing.  Super subtle.

andamian:
> I think it's fine without it. People can make their own lightweight conversion methods to do what they want if they need to. Also great fix on that doc thing. Super subtle.

doc errors are annoying because they don't provide much context (or I don't know how to make them do)...

funbaker:
About that time format (not related to this PR): It would be good to have this format in the astropy datetime class at some point.

andamian:
Thank you both for your contributions. What else is there for 1.0? Maybe we should go through the list of issues and label those required for this release. Or some priority order, maybe.

PR #172, opened 2019-08-29
Update ex_casA_image_cat example

### LINKED ISSUE #170: ex_casA_image_cat example doesn't return any results
After using SkyCoord, I noticed that the ex_casA_image_cat.py example still doesn't return anything.

Everything runs, but it looks like it finds no archives based on the parameters it is given:

```
pyvo.regsearch(servicetype='image', waveband='xray')
```

Returns an empty table.  I'm guessing it did return things at some point.  If I get rid of the waveband parameter, it returns a whole bunch.  I'm not sure how the example should change, but it's somewhat bad form to have an example that doesn't return results.

cbanek:
Update ex_casA_image_cat example

Thanks @msdemlei for the updated working example.
    
Taken verbatim from #170. Fixes #170.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,8 @@
 
 - Stop installing files from scripts to /usr/local/bin. Move them to examples/images instead. [#166]
 
+- Update ex_casA_image_cat example. [#172]
+
 0.9.3 (2019-05-30)
 ==================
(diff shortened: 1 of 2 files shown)


cbanek:
This is a branch off of #166

PR #193, opened 2019-11-12
Fix docstring build error

cbanek:
This is a simpler suggestion to fix the build error, which has been
suggested on #189 and also https://github.com/sphinx-doc/sphinx/issues/6753

By having this empty init comment, this should get the build going
green again.


pyvo/io/uws/tree.py
```diff
@@ -278,6 +278,7 @@ class Parameters(UWSElement, HomogeneousList):
     Parameters element of a job
     """
     def __init__(self, config=None, pos=None, _name='parameters', **kwargs):
+        """ """
```
andamian: I would add a line below to state that the empty `pydoc` is required and should not be removed. Otherwise someone might be tempted to get rid of it and they'll scratch their heads trying to figure out why sphinx is unhappy. I can picture myself in that situation :-).

cbanek:
Okay sounds good, thanks for approving that.  I hit a bump with some kind of network disconnection in the previous travis run, so I bumped it, and all looks well.  Going to merge this.

PR #194, opened 2019-11-13
Get auth docs building

cbanek:
For some reason, the docstrings for the auth classes aren't
being built into the docs.  Add a link from the main page to
get things generated.

### DIFF
--- a/pyvo/auth/__init__.py
+++ b/pyvo/auth/__init__.py
@@ -1,1 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+__all__ = ["AuthSession", "AuthURLs", "CredentialStore"]
+
+from .authsession import AuthSession
+from .authurls import AuthURLs
+from .credentialstore import CredentialStore
--- a/docs/auth/index.rst
+++ b/docs/auth/index.rst
@@ -7,6 +7,10 @@ communicating with virtual observatory services.
 Reference/API
 =============
 
+.. automodapi:: pyvo.auth
+
 .. automodapi:: pyvo.auth.authsession
+
 .. automodapi:: pyvo.auth.authurls
+
 .. automodapi:: pyvo.auth.credentialstore
--- a/docs/auth/index.rst
+++ b/docs/auth/index.rst
@@ -1,0 +1,12 @@
+******************
+Auth (`pyvo.auth`)
+******************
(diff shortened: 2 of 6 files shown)


cbanek:
So here's the build error that has me perplexed:

/home/travis/build/astropy/pyvo/docs/auth/index.rst.rst:19:autosummary: stub file not found 'pyvo.auth.authsession.AuthURLs'. Check your autosummary_generate setting.
Sphinx Documentation subprocess failed with return code 2

So, it should be pyvo.auth.authurls.AuthURLs is the python import path for that module, not authsession.  I'm not sure where it's getting this path or how it is generating it, but obviously I am doing something wrong here.

jonathansick:
At the top of `authsession.py`, add this:

```
__all__ = ('AuthSession',)
```

What's going on is that automodapi things `AuthURLs` and `CredentialStore` are defined in `authsession.py` because they're imported into the namespace for that module. Using the `__all__` curates that namespace to just the stuff that the authsession module itself defines.

You might also want to import `AuthSession`, `AuthURLs`, and `CredentialStore` into the `__init__.py` of `pyvo.auth` and then have one automodapi directive pointing at that namespace — but that's an API design decision.

pyvo/auth/__init__.py
```diff
@@ -1 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+__all__ = ["AuthSession", "AuthURLs", "CredentialStore"]
+
+from .authsession import AuthSession
```
andamian: I usually `import *` here so when new things are added in the submodule all I need to do is update `__all__` for that module.

funbaker: https://stackoverflow.com/questions/2386714/why-is-import-bad

andamian: This is the only case when I ever use `import *`. The usual pitfalls do not apply here because  it only pulls what it is in the defined `__all__` of the modules and there is no code in `__init__`. 

I prefer `__all__` in the submodule as it's closer to the actual code and easier to get the context but I don't have a strong opinion.

funbaker: Mea culpa. In that case it should be fine, explicitness is ensured...

andamian:
@cbanek Merged this - finally. Sorry for the delay.

PR #197, opened 2019-12-12
Handle description of None when describing a TAP service's tables

stargaser:
This small PR is to address an issue we have found in describing available tables from a TAP service, when a description is not provided.

To test:
```
import pyvo as vo
irsa_services = vo.regsearch(servicetype='table', keywords=['irsa'])
irsa_tap = irsa_services[0]
irsa_tap.service.tables.describe()
```
The fix in the PR is to print `No description` instead of raising an exception.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@
 
 - Add default for UWS version. [#199]
 
+- Handle description of None when describing a TAP service's tables. [#197]
 
 1.0 (2019-09-20)
 ================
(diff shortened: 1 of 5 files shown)


cbanek:
Since this is a user facing kind of thing, might want to add an entry to the CHANGELOG (although I'm also fine if we think this is below that bar, we can add the flag no-changelog-entry-needed)

stargaser:
I feel like I should add a test. Looks like I should make an analog of `tests/data/tables/multiple_table_descriptions.xml` that is `missing_table_description.xml` and then add a test using that file. Does that make sense?

stargaser:
Cross-referencing NASA-NAVO/workshop-dev#5 where we noticed the problem addressed in this PR.

stargaser:
Huh, looks like codecov/project increased by adding the test, but codecov/patch went down. I'm puzzled.

Issue #198, opened 2019-12-17
UWS fails on missing version attribute


tomdonaldson:
When reading a UWS response with a `<job>` element that has no `version` attribute, as can happen with asynchronous queries on some TAP services, this exception is raised:
```
AttributeError: 'LooseVersion' object has no attribute 'version'
```
Although that `version` attribute is required for UWS version 1.1 services, it remains optional for UWS v1.0 (see [the UWS 1.1 specification](http://www.ivoa.net/documents/UWS/20161024/REC-UWS-1.1-20161024.html#versioning) section on versions).  So if the UWS response is missing the `version` attribute, version 1.0 should be assumed.

tomdonaldson:
Fixed with the merge of #199

PR #199, opened 2019-12-17
Add default for UWS version - Issue #198

tomdonaldson:


pyvo/io/uws/endpoint.py
```diff
@@ -108,7 +108,9 @@ def parse(self, iterator, config):
             if start and tag == 'xml':
                 pass
             elif start and tag == 'job':
-                self._version = data.get('version')
+                # Default to v1.0 since the version attribute
+                # was not required until version 1.1.
+                self._version = data.get('version', "1.0")
```
stargaser: Minor stylistic point that single quotes and double quotes are used in the same line of code.

tomdonaldson: I made the requested changes and squashed the commits.

stargaser:
🎉

Issue #217, opened 2020-02-26
For TAP or SIA results, support DataLink queries with multiple IDs


andamian:
Current implementation requires separate calls to DataLink for each of the records in the results returned by an SIA or TAP query. In practice, this sometimes proves inefficient.

I propose an optimization by which the API allows the user to send multiple IDs (batches) to the DataLink service. In our experience, the overhead for individual calls is too high. I suspect that the new functionality will be in the `DatalinkResultsMixin` by potentially adding a batch size argument maybe?

andamian:
@funbaker - please let me know your opinion on this. What I'm after is the ability to retrieve datalink info in batches rather than individual calls for each record. It doesn't seem to be supported at the moment. Please let me know if you have a suggestion better than the one I've proposed. Tomorrow, I'm attempting an implementation. Thanks

msdemlei:
On Wed, Feb 26, 2020 at 01:53:08AM -0800, andamian wrote:
> Current implementation requires separate calls to DataLink for each
> of the records in the results returned by an SIA or TAP query. In
> practice, this sometimes proves inefficient.
> 
> I propose an optimization by which the API allows the user to send
> multiple IDs (batches) to the DataLink service. I suspect that the
> new functionality will be in the `DatalinkResultsMixin` by
> potentially adding a batch size argument maybe?

This was certainly the intent of making ID repeatable in datalink; I
have a hunch this will become a bit tedious with associated
services, but I'd hope at least DaCHS services will do the right
thing for multiple ids.  You could try this with matches from, for
instance, the SSA service at
http://dc.zah.uni-heidelberg.de/feros/q/ssa/ssap.xml?

Other than that, I'm all for exposing multi-ID in pyVO just about
now, because right now no client I'm aware of does multi-ID, and that
means that related bugs in services will accumulate over time.

andamian:
@msdemlei thanks. we've been using multi IDs with the cadc DataLink service just not through `pyvo`.  Good to know that I have another service to test against.

PR #218, opened 2020-02-27
Batched DataLink calls

andamian:
#217


andamian:
First attempt. I've tried to make it a drop in change, transparent change. A bit tricky since `DatalinkResults` was designed to hold entries for a single ID. The work around was to cache multiple calls to datalink but return only rows that belong to a specific ID. Not that clean, I admit but it seems to work. I kept the old `DatalinkResultsMixin.iter_datalinks` method for now just a bench mark for correctness and performance but it can be replaced by the corresponding `iter_datalinks` without any effect (hopefully) on the calling code. 
@msdemlei & @funbaker or anyone else, please let me know if you can suggest a better approach. I wouldn't leave the decision on the batch size on the user, but the variable can be tweaked. This is the first version so it's not entirely optimized but it works with a few real examples I've tried.

msdemlei:
On Wed, Feb 26, 2020 at 08:59:36PM -0800, andamian wrote:
> @msdemlei & @funbaker or anyone else, please let me know if you can
> suggest a better approach. I wouldn't leave the decision on the
> batch size on the user, but the variable can be tweaked. This is
> the first version so it's not entirely optimized but it works with
> a few real examples I've tried.

I suppose given the iter_datalinks approach this is about how it
should be done.  There is a bit of a problem with this approach
because datalink services can have match limits.  10000 for such a
limit is plausible, and 10 datalinks per id is also not unheard-of (I
have some datasets with ~100 datalinks).

Hence, for DATALINK_BATCH_SIZE=1000 you might well run into a match
limit.  That, I think, shouldn't pass unnoticed.  The way the spec is
written, it's relatively simple to catch such a situation: After
doing the batched query, you could do the equivalent (I've not lookad
at what sort of thing _get_datalinks actualy returns) of

```
  if set(ids)!=set(r["ID"] for r in self._cached_dl):
    raise WhateverError("Datalink batched query overflowed match limit;"
      " Decreate DATALINK_BATCH_CALL_SIZE.")
```

To make the code a bit more transparent, I'd pull out the cache
filling into a separate method.

You can also pass MAXREC to your datalink call, but since we have no
way of determining the hard limit on the service at this point, I
don't think that's useful until we've thought a bit more about
additional service metadata (which brings us back to caproles...).

       -- Markus

andamian:
@msdemlei the batch size is in the input but the server limitation is in the output. The client can't come up with the optimal batch size since it doesn't know how many data links it would generate. Each different call with the same batch size can generate a different number of output rows.

I was thinking of a different approach: send all the IDs at once, and if the server returns OVERFLOW status remove the processed ones from the list before sending them again. Continue until status OK. Do you think this will put unnecessary burden on the service?
To accommodate large number of IDs we might need to change the submission from the current GET to a POST but I think the spec supports that.

What do you think?

andamian:
BTW, one of the scenarios that astronomers have asked is:
1. Perform SIA2 query
2. Process results which typically is just some further filtering that cannot be achieved with the query.
3. Download data
How can step 2 be performed with the `SIAResults` object? The modified `SIAResults` object is required in step 3 since it stores information about the DataLink step. Users should be able to easily perform record deletes from an `SIAResults`, no? Maybe the functionality is there but I'm missing it.

msdemlei:
On Thu, Feb 27, 2020 at 11:45:51AM -0800, andamian wrote:
> I was thinking of a different approach: send all the IDs at once,
> and if the server returns OVERFLOW status remove the processed ones
> from the list before sending them again. Continue until status OK.
> Do you think this will put unnecessary burden on the service?

I think that's a totally reasonable thing to do.  If you're going to
iter_datalinks fairly exhaustively anyway, the server will certainly
be grateful if you're firing off a few large requests rather than a
couple of thousand small requests.

Perhaps one needs to think about really large queries, as in several
1e5 or so; a datalink record with an attached service block may be a
few k, so the results may work out to be in the Gigabyte range.
But then that should probably be done on the server side by
introducing a reasonable match limit there.

> To accommodate large number of IDs we might need to change the
> submission from the current GET to a POST but I think the spec
> supports that.

It should definitely be POST either way; Datalink sect. 2.1 requires
support for both, and the only reason I see for having GET in the
first place is so people can pass around canned queries as URL
literals.  For dynamic queries, POST is a lot more robust all around.

andamian:
@msdemlei @funbaker I've pushed another version. Initially, the client sends all the IDs with a POST. If the result is incomplete, the size of the returned result represents the size of the batch. Subsequent calls are initiated until all the IDs are resolved.
Please review at your earliest convenience. Thanks

andamian:
Fixed tests broken due to upstream changes (https://github.com/astropy/astropy/pull/9505 )

pyvo/dal/adhoc.py
```diff
@@ -491,14 +594,14 @@ def get_first_proc(self):
...
-    Mixin for soda functionallity for record classes.
+    Mixin for soda functionality for record classes.
     If used, it's result class must have
     `pyvo.dal.datalink.AdhocServiceResultsMixin` mixed in.
     """
     def _get_soda_resource(self):
         try:
             return self._results.get_adhocservice_by_ivoid(
-                b"ivo://ivoa.net/std/SODA#sync")
+                SODA_SYNC_IVOID.encode('utf-8'))
```
msdemlei: I'd like the interface better if get_adhocservice_by_ivoid accepted strings and did the legacy encoding (where necessary) itself.

andamian: Agree and fixed.

pyvo/dal/adhoc.py
```diff
@@ -154,12 +168,68 @@ class DatalinkResultsMixin(AdhocServiceResultsMixin):
...
+                            'Could not retrieve datalinks for: {}'.format(
+                                ', '.join([_ for _ in
+                                           self._remaining_ids])))
+                    self._batch_size = len(self._current_ids)
+                id = self._current_ids.pop(0)
+                self._processed_ids.append(id)
+                self._remaining_ids.remove(id)
+                yield self._current_batch.clone_byid(id)
+            else:
+                yield DatalinkResults.from_result_url(row.getdataurl())
```
funbaker: Umm. No. Please stick to `getdatalink` because of parameters which are described in the `RESOURCE` element.

pyvo/dal/adhoc.py
```diff
@@ -154,12 +168,68 @@ class DatalinkResultsMixin(AdhocServiceResultsMixin):
...
-        for record in self:
-            yield record.getdatalink()
+        # To reduce the number of calls to the Datalink service, multiple
+        # IDs are sent in batches. The appropriate batch size is not available
+        # as each service has its own criteria for determining the maximum
+        # number of IDs processed per request. To overcome this limitation,
+        # this method initially sends all the available IDs and if the
+        # response is partial, uses the size of the response as the batch
+        # size.
+        if not hasattr(self, '_datalink'):
```
funbaker: As this is a generator, wouldn't it be better to turn those attributes into local variables?

pyvo/dal/adhoc.py
```diff
@@ -154,12 +168,68 @@ class DatalinkResultsMixin(AdhocServiceResultsMixin):
...
+                self._datalink = None
+        if not hasattr(self, "_current_batch"):
+            # static vars
+            self._remaining_ids = []  # remaining IDs to processed
+            self._current_batch = None  # retrieved but not returned yet
+            self._current_ids = []  # retrieved but not returned
+            self._processed_ids = []  # retrived and returned IDs
+            self._batch_size = None  # size of the batch
+
+        for row in self:
```
funbaker: About that readabillity...
It's fine for now though, a lot of comments.

pyvo/dal/adhoc.py
```diff
@@ -154,12 +168,68 @@ class DatalinkResultsMixin(AdhocServiceResultsMixin):
...
+            if self._datalink:
+                if not self._current_ids:
+                    if self._batch_size is None:
+                        # first call.
+                        self.query = DatalinkQuery.from_resource(
+                            [_ for _ in self], self._datalink)
```
funbaker: I'm not happy about this parameter wrangling.
Maybe modify `from_resource` to have keyword parameters for different types of input?
The name is fine as the resource is always passed.

pyvo/dal/adhoc.py
```diff
@@ -323,6 +397,8 @@ def from_resource(cls, row, resource, session=None, **kwargs):
         for name, query_param in kwargs.items():
             try:
                 input_param = find_param_by_keyword(name, input_params)
+                if input_param and query_param is None:
```
funbaker: Good catch, this could have been a source of subtle bugs.

pyvo/utils/__init__.py
```diff
@@ -0,0 +1 @@
+from .compat import *
```
funbaker: Missing newline before EOF

trjaffe:
This doesn't work for the HEASARC TAP service's DataLinks.  The reason is that our service uses the integer row number as the ID.  So in the VOTable, there's:

>    \<FIELD ID="DataLinkID" datatype="int" name="__row">
> ...
>  \<RESOURCE type="meta" utype="adhoc:service">
>   \<PARAM ID="standardID" arraysize="*" datatype="char" name="standardID" value="ivo://ivoa.net/std/DataLink#links-1.0"/>
>   \<PARAM ID="accessURL" arraysize="*" datatype="char" name="accessURL" value="https://heasarc.gsfc.nasa.gov/xamin/vo/datalink/chanmaster"/>
>  \</RESOURCE>


Then in adhoc.py:213, the following code

`                    self._current_ids = list(OrderedDict.fromkeys(
                        [_ for _ in self._current_batch.to_table()['ID']]))
`
gets the IDs from the DataLinkResults as a list of strings, while the self._remaining_ids are from the TAPResults votable as integers.  So then the remove() further down doesn't work.  

So Tom McGlynn sent a message to the DAL group asking about this, since the standard document doesn't make clear that IDs have to be strings.  So which needs to be fixed, our TAP service to use datatype="char" for the row or PyVO to convert in the above check?  I found two places that a simple type conversion would fix this in PyVO.  (The other is adhoc.py:556 for the same reason.)  Of course, it could also be fixed in our service.  

Thoughts?

funbaker:
summoning @msdemlei for a standard related question

msdemlei:
On Wed, Apr 01, 2020 at 10:30:18AM -0700, trjaffe wrote:
> So Tom McGlynn sent a message to the DAL group asking about this,
> since the standard document doesn't make clear that IDs have to be
> strings.  So which needs to be fixed, our TAP service to use
> datatype="char" for the row or PyVO to convert in the above check?
> I found two places that a simple type conversion would fix this in
> PyVO.  (The other is adhoc.py:556 for the same reason.)  Of course,
> it could also be fixed in our service.  
> 
> Thoughts?

I'm not so sure about the "simple type conversion" -- type
conversions have a way of eventually turning out hairy.

Anyway, this is clearly a lacuna in the specs; actually, both VOTable
(which is hardly to blame) and Datalink, so I don't think we can
simply decide this here.  My current position is in

http://mail.ivoa.net/pipermail/dal/2020-April/008322.html

-- were I don't think any amount of discussion will change my stance
on avoiding different type/arraysize/xtype tuples on the two ends of
a ./@ref relationship.

funbaker:
I see two possibilities:

1. make everything `char`
2. do implicit conversion but raise a warning if it's with non-trivial types in terms of conversion.

pyvo/dal/adhoc.py
andamian: I don't see any harm in having `_datalink` (pointer to service) as a global for `DatalinkResult`. It is an attribute to that response.

pyvo/dal/adhoc.py
andamian: How would you define the argument of `from_resouce`? Need to keep `row` for backwards compatibility and have an `additional_rows` one? I just cant think of a better way. Which I can rename `row` to `rows`. Please let me know if you have a better option - I'm open to it.

pyvo/dal/adhoc.py
andamian: That was wrong. I've replaced it with: `row.access_format == 'application/x-votable+xml;content=datalink'` (according to the specs)

pyvo/utils/__init__.py
andamian: Fixed

andamian:
> A belated update: the DAL group is going to clarify the standard that the ID should be char. We haven't yet finished fixing our service, but this mod works with others and looks fine to everybody else. Nor did I have any other comments. So I think this PR is good to go.

Thank you @trjaffe for looking into it.

pyvo/dal/adhoc.py
funbaker: `_datalink` as a attribute of the result may be fine, but not `_current_batch`. When someone creates two generator instances for example.

pyvo/dal/adhoc.py
funbaker: `rows` sounds good indeed.

pyvo/dal/adhoc.py
andamian: Doh! You are perfectly right! I've missed that somehow - but fixed now. Very good point.

pyvo/dal/adhoc.py
andamian: Renamed to `rows`. Existing usages of the form `from_resouce(some_row, some_resource)`  would still work but not `from_resource(row=some_row, resource=some_resource)`. If we want to make that backwards compatible I would need to have defaults for both `rows` and `resource` and check for calls with `row` in `kwargs`. Too high of a price to pay I'm thinking, no?

pyvo/dal/params.py
```diff
@@ -387,16 +391,21 @@ def get_dal_format(self, val):
...
             if not isinstance(low, Quantity):
                 low = low*self._unit
             low = low.to(self._unit, equivalencies=self._equivalencies).value
             if not isinstance(high, Quantity):
                 high = high*self._unit
             high = high.to(self._unit, equivalencies=self._equivalencies).value
-        if low > high:
-            raise ValueError('Invalid interval: min({}) > max({})'.format(
-                low, high))
+
```
andamian, line 405: @funbaker pardon for these unrelated changes I'm trying to get in with this PR. I need them for work in `astroquery.alma`

pyvo/dal/adhoc.py
funbaker: I don't think this should be a problem. At least i don't think anyone is using `from_resource` directly. Hmm.

pyvo/dal/params.py
funbaker: Not a very clean practice to do unrelated changes in the same PR...
Not even a explanation what this is doing exactly?

This way it is not easy to revert, no one cane figure out why this is here, etc.

andamian: Pull out. Now part of #225

andamian:
@funbaker - I think I've now addressed all the concerns. Please let me know if that's not the case. Thanks

pyvo/dal/adhoc.py
funbaker: Hmm. I think there should be an `else` clause, even if it just emits warnings.

pyvo/dal/adhoc.py
```diff
@@ -154,12 +171,57 @@ class DatalinkResultsMixin(AdhocServiceResultsMixin):
...
+                        # we are done
+                        return
+                    if batch_size:
+                        # subsequent calls are limitted to batch size
+                        self.query['ID'] = remaining_ids[:batch_size]
+                    current_batch = self.query.execute(post=True)
+                    current_ids = list(OrderedDict.fromkeys(
+                        [_ for _ in current_batch.to_table()['ID']]))
+                    if not current_ids:
+                        raise DALService(
```
funbaker, line 215: This one was easy to miss - it should be `DALServiceError`

andamian: Indeed. Good eye. Double checked and this was the only occurrence of the typo.

pyvo/dal/adhoc.py
andamian: `else: yield None` maybe since there is no info to determine the corresponding datalink for that row?

funbaker: tbh I'm not sure whats the best way here. There shouldn't be empty rows but they also shouldn't go unnoticed.

andamian: The iterator is returning Datalink resources associated with each row. Question is what to do when no such resources are available (access_url is probably a direct url in that case). This case is valid so I don't think the method should err or warn. The question IMO is whether to return `None` or continue and now that I'm thinking more about it I'm more inclined for the latter (`None` in itself is not a useful info). The contract of the method is to return corresponding Datalink resources so skipping the rows that don't have such resources is OK, no?

This is the relevant section of SIA spec:
```3.1.3 Use of access_url and access_format
If the SIA service is only dealing with simple data (one file per result), the 
access_url column may be a link directly to that file, in which case the 
access_format column should specify the file format (e.g. application/fits).

If the data provider implements a DataLink service for the data being found 
via the SIA {query} capability, they may put a URL to invoke the DataLink 
{links} capability (with ID parameter and value) in the access_url column; if 
they do this, they must also put the standard DataLink MIME type [9] in the 
access_format column.```

funbaker:
> @funbaker - I think I've now addressed all the concerns. Please let me know if that's not the case. Thanks

Aye, should be fine for now.

pyvo/dal/adhoc.py
andamian: So, better to return `None` instead of `continue` you think?

Issue #221, opened 2020-04-18
Support of custom parameters missing in sia2


andamian:
This is an oversight on my side. SIAv2 has support for custom parameters so the library should support it too.

PR #222, opened 2020-04-18
Sia2 kwargs

andamian:
#221

### DIFF
--- a/pyvo/dal/sia2.py
+++ b/pyvo/dal/sia2.py
@@ -84,7 +84,7 @@ max_records : int
     allows the client to limit the number or records in the response
 **kwargs : custom query parameters
     single or a list of values (or tuples for intervals) custom query
-    parameters that a specific service accepts. The values of the 
+    parameters that a specific service accepts. The values of the
     parameters need to follow the SIAv2 format and represent the
     appropriate quantities (where applicable).
 """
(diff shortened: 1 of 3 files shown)


cbanek:
Guessing the build break is some kind of upstream future votable thing in astropy?

tomdonaldson:
> Guessing the build break is some kind of upstream future votable thing in astropy?

Yes, @cbanek .  Surely this is due to https://github.com/astropy/astropy/pull/9505, which will be released soon with astropy 4.1.

Whether these failures get fixed here to pass with both `byte` and `str` values, we should probably think about whether we have other coordination to do with respect to that change.  I.e., which pyvo versions should pass testing with both astropy 4.0 and 4.1?

andamian:
All green now! @funbaker do you have time to take a look? Thanks

pyvo/dal/sia2.py
```diff
@@ -305,6 +312,16 @@ def __init__(self, url, pos=None, band=None, time=None, pol=None,
         for rf in _tolist(res_format):
             self.res_format.add(rf)
 
+        for ii in kwargs:
```
funbaker, line 315: `for name, value in kwargs.items()` would be more readable

andamian:
@funbaker - anything else needed for the approval?

PR #226, opened 2020-05-04
Add license file

aragilar:
I was surprised that pyvo didn't have a license file, I've added this which is similar to the format used by other astropy projects.


andamian:
@aragilar - Merged. Thank you for your contribution.

PR #227, opened 2020-05-08
MNT: Ignore import error in pytest collection

pllim:
This is an automated update made by the `batchpr` tool :robot: to future-proof test collection behavior `pytest-doctestplus` - feel free to close if it doesn't look good or isn't relevant! You can report issues to @pllim.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,3 +45,6 @@ install_requires = astropy requests mimeparse
 version = 1.1.dev
 
 [entry_points]
+
+[tool:pytest]
+addopts = --doctest-ignore-import-errors


andamian:
@pllim, without some context it's hard to make the call. What prompted this update?

pllim:
Ah, sorry. Maybe I should have explained more above. This mainly affects people who already use `pytest-doctestplus`. Currently, it ignores import errors as if this option is already activated. But it is not the default behavior in upstream `doctest`, where this option is explicitly given. In the future, if this hack below is reverted, you will need this option. But all that is only relevant if you already or plan to use `pytest-doctestplus`.

https://github.com/astropy/pytest-doctestplus/blob/dfc9e9f05f8c63f003d23fb82deacec11c6e3a11/pytest_doctestplus/plugin.py#L168-L176

andamian:
I'm not sure it is used in `pyvo`. I see it installed but maybe it's pulled by `pytest-astropy`. Unless someone knows for sure we use it, I'd say we keep this PR around in case the build brakes when upstream changes are released and when/if this happens we can merge the change.

pllim:
Maybe you don't need it.

PR #236, opened 2020-06-26
Begin 1.2 release cycle

cbanek:


cbanek:
Sorry I think this got forgotten.  We should probably take this first before we merge the dropping python part?

andamian (approved):

Issue #231, opened 2020-06-16
datalink: bysemantics should by default include the entire branch


msdemlei:
In case you're not aware of datalink in pyVO: there's examples in http://docs.g-vo.org/pyvo/html/page026.html.

In datalink, rows have an associated semantics; the terms there come from the controlled, hierarchical vocabulary at http://www.g-vo.org/rdf/datalink/core.  As you can see, there is structure in there; for instance, bias is a child of computation.

Now, I would like get bysemantics method of datalink results to take into account this structure by including, by default, all narrower terms, too.  I'd like to put this is before any real code depends on the current, IMHO less desirable behaviour.  I'd volunteer to implement that, but I'd be grateful for a piece of advice (below).

Consider, as a (non-ideal) example, http://dc.g-vo.org/kapteyn/q/dl/dlmeta?ID=ivo%3A//org.gavo.dc/~%3Fkapteyn/data/fits/POT015_000317.fits.  As you can see, this has two sorts of previews (the details don't really matter).  Now, if you run:

```
from pyvo.dal import adhoc

links = adhoc.DatalinkQuery(
9"http://dc.g-vo.org/kapteyn/q/dl/dlmeta",
9id="ivo://org.gavo.dc/~?kapteyn/data/fits/POT015_000317.fits").execute()
print(list(links.bysemantics("#preview-image")))
print(list(links.bysemantics("#preview")))
```

you will see that the second list (for the wider term) does not include
the first list, although of course everything in the first list "is-a"
preview by the semantics.  I would like to fix this; the second list
should include the first.

Concretely, I'm proposing that the bysemantics method of `dal.adhoc.DatalinkResults` should grow a keyword parameter `exact`, defaulting to `False`; and if `exact` is false, all terms narrower than the first argument should be included in the filter (there's a bit of legalese that needs to restrict this to terms in the IVOA vocabulary and such, but that's minor details).

Given recent developments in IVOA semantics, that's rather simple to
accomplish: You pull a file like
http://www.ivoa.net/rdf/datalink/core/2020-02-12/datalink.desise,
json-decode it and pull out the narrower terms for what you already
have (details on
http://ivoa.net/documents/Vocabularies/20200612/WD-Vocabularies-2.0-20200612.html#tth_sEc3).

[...]

PR #241, opened 2020-09-30
Make datalink .bysemantics semantics-aware

msdemlei:
This will fix bug #231.

Specifically, this PR adds utils.vocabularies, which is usable beyond
datalink to pull IVOA vocabularies as specified by the (current)
Vocabularies in the VO 2 WD (I will update the module in case there
are major changes to the WD),
http://ivoa.net/documents/Vocabularies/20200612/ .

Using this, it extends DatalinkResults.bysemantics to accept lists as
well as strings (as before).  It also accepts "naked" terms now (i.e.,
without the hash that is a bit of a serialisation hack in datalink).
I'm not attempting to recognise full resource URIs at this point (but
that would be easy and safe to add once someone actually has a use for that).

Most importantly, as long as include_narrower is True (the default),
this consults the datalink vocabulary and will match narrower terms as
well; for instance, just looking for calibration will give you dark, bias, and flat as
well.


andamian:
@msdemlei - We've reworked the CI. Please pull the master to see if that fixes the tests so that we can proceed with the next steps. And sorry for the delay.

msdemlei:
On Wed, Dec 09, 2020 at 08:00:27AM -0800, andamian wrote:
> @msdemlei - We've reworked the CI. Please pull the master to see if that fixes the tests so that we can proceed with the next steps. And sorry for the delay.

Hm -- regrettably not quite.

The changelog entry I've added (but not pushed yet for environmental
concerns).  The failures with astropy 4.1 I've not quite figured out
yet, mainly because I got lost in the pytest version hell.  With
pytest 6.1.2 in the virtual environment (on top of Debian 3.7) the
thing consistently throws

  cannot import name 'transfer_markers' from '_pytest.python'

when trying to run any tests.  Before I delve into this any deeper:
Is there a known solution for this?

andamian:
@msdemlei I took the liberty of fixing the tests. Small change - they are now passing. Please pull.

Sorry, I'm not familiar with the `pytest` problem on `Debian`. If you are familiar with `tox`, you can let `tox` deal with virtual environment setup. All the tests in CI rely on that, so if it's something wrong with it they can all benefit.

Feel free to push the change long to this branch. That should clear the other remaining error.

Sorry for the duplicates in the tests - the CI framework was created mainly for PR from forks and I'm not quite sure how to fix it when the PR originates from the master repo.

msdemlei:
Can I bump this?  I'm not really sure if the failed codecov tests are something I should be trying to fix of if that's just CI noise...

andamian:
The report says that the coverage decreased slightly. If you can bump the coverage for your patch from 62% to 75%, it will pass. If you follow the `Details` link in `codecov/patch` it highlights the new code that is not tested. Maybe you could add a test or two to push it over the 75% and turn everything green. The way is set right now, it doesn't look like GH allows merging otherwise.

msdemlei:
On Thu, Mar 04, 2021 at 09:44:22AM -0800, andamian wrote:
> The report says that the coverage decreased slightly. If you can
> bump the coverage for your patch from 62% to 75%, it will pass. If
> you follow the `Details` link in `codecov/patch` it highlights the
> new code that is not tested. Maybe you could add a test or two to
> push it over the 75% and turn everything green. The way is set
> right now, it doesn't look like GH allows merging otherwise.

Regrettably, the codecov links all give me "403 unauthorised", and my
patience for debugging and working around crazy auth schemes "in the
cloud" is relatively low, in particular where codecov.io's docs don't
work unless you let them execute Javascript (WTF?).

But I'd guess the reason coverage looks so bad is that
utils.vocabularies has the conditional code for astrpy versions <4
that won't be exercised in version 4 because it simply doesn't run in
that version.

It turns out codecov use coverage.py, and so I'm escaping from this
by just using a pragma to keep the compatibility code out of
coverage.py's view.  I'd say that's fair, even more so since we can
perhaps drop the compatibility code soon-ish anyway (though I'm
always a big fan of maintaining backwards compatiblity as long as we
can).

After that, the codecov is fine with the change.

andamian:
I'm not sure why that info is not accessible. Maybe @pllim might have an idea.

The coverage report is also available from the "Checks" tab as GH code comments.

Is the PR now ready for code review? Do you have a suggestion for a reviewer?

pllim:
Hello. Exactly which link gave you a 403? Can you please clarify?

I know CircleCI now requires login to see the results but I am not sure about Codecov.

msdemlei:
On Fri, Mar 05, 2021 at 09:41:11AM -0800, andamian wrote:
> I'm not sure why that info is not accessible. Maybe @pllim might have an idea.

Hm... the equivalent links the github PR page now shows work.
Perhaps the previous results were just expired?

> Is the PR now ready for code review? Do you have a suggestion for a reviewer?

I'd like to think it could be reviewed, but I'm afraid I can't
suggest a reviewer.  On the other hand, I'd say the code as such by
and large is straightforward, so no special skills are required
there.  As to the feature added itself, perhaps the original issue
helps giving the larger picture?

andamian:
In this case, I'll do the code review. @cbanek or @funbaker or anyone else interested please feel free to jump in.

funbaker:
Gonna try to have a look at the weekend.

pyvo/utils/vocabularies.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def get_label(voc, term):
+    """returns the label of term if it's in the desise vocabulary voc,
+    term capitalised otherwise.
+    """
+    if term in voc["terms"]:
+        return voc["terms"][term]["label"]
+    else:
+        return term.capitalize()
```
andamian, line 98: Why capitalize? Probably something that I'm not familiar with in dictionaries. Also, shouldn't this trigger a refresh of the cache or it's not worth it?

msdemlei:
On Fri, Mar 12, 2021 at 04:09:57PM -0800, andamian wrote:
> +def get_label(voc, term):
> +    """returns the label of term if it's in the desise vocabulary voc,
> +    term capitalised otherwise.
> +    """
> +    if term in voc["terms"]:
> +        return voc["terms"][term]["label"]
> +    else:
> +        return term.capitalize()
> 
> Why capitalize? Probably something that I'm not familiar with in
> dictionaries. Also, shouldn't this trigger a refresh of the cache
> or it's not worth it?

The capitalisation is just a matter of taste and could be dropped --
on the other hand, human-readable labels should generally be
capitalised in the vocabularies, and doing it for the occasional
out-of-vocabulary term probably improves general apprearance.  Still,
I'll drop it if you think it's unneccessary frills.

On refreshing: I'm pretty sure a cache miss shouldn't trigger
updates, as cache misses could come with high frequency; you'd then
have to have some sort of rate limiting ("only re-try after N
hours"), and at least at this point this seems a lot more
complication than it might one day help.

Generally, I'd say any auto-updating policies should be developed as
we gain experience with how vocabularies get to be used.

andamian:
> On Fri, Mar 12, 2021 at 04:09:57PM -0800, andamian wrote: +def get_label(voc, term): + """returns the label of term if it's in the desise vocabulary voc, + term capitalised otherwise. + """ + if term in voc["terms"]: + return voc["terms"][term]["label"] + else: + return term.capitalize() Why capitalize? Probably something that I'm not familiar with in dictionaries. Also, shouldn't this trigger a refresh of the cache or it's not worth it?
> The capitalisation is just a matter of taste and could be dropped -- on the other hand, human-readable labels should generally be capitalised in the vocabularies, and doing it for the occasional out-of-vocabulary term probably improves general apprearance. Still, I'll drop it if you think it's unneccessary frills. On refreshing: I'm pretty sure a cache miss shouldn't trigger updates, as cache misses could come with high frequency; you'd then have to have some sort of rate limiting ("only re-try after N hours"), and at least at this point this seems a lot more complication than it might one day help. Generally, I'd say any auto-updating policies should be developed as we gain experience with how vocabularies get to be used.

If capitalization is desired, should we try to do it even for the labels of existing terms? That aside, I would suggest adding a `default` argument to the method that is returned when the term no not found. That would be consistent with dictionary functionality.

OK on refreshing.

msdemlei:
On Mon, Mar 15, 2021 at 02:47:37PM -0700, andamian wrote:
> If capitalization is desired, should we try to do it even for the
> labels of existing terms? 

No -- the vocabulary designers may have good reasons for using lower
case in labels (think of astronomical nomenclature that is often
case-sensitive).  On the other hand, when people invent terms,
there's probably little thought involved :-).  But as I said: if
you're worried, I'll remove it.

> That aside, I would suggest adding a `default` argument to the
> method that is returned when the term no not found. That would be
> consistent with dictionary functionality.

That is a good point.  I've put it in in 3749b7c2.

pyvo/utils/vocabularies.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def get_label(voc, term, default=None):
+    """returns the label of term if it's in the desise vocabulary voc,
+    term capitalised otherwise.
+    """
+    if term in voc["terms"]:
+        return voc["terms"][term]["label"]
+    else:
+        return default if default is not None else term.capitalize()
```
andamian, line 98: I would go further and just return `default`. Especially if default capitalized term is more a matter of taste as you said. Right now it hides the case when the label does not exist. But I'm fine with this too.

msdemlei:
On Wed, Mar 17, 2021 at 08:45:16PM -0700, andamian wrote:
> @andamian commented on this pull request.
> 
> 
> 
> +    else:
> +        return default if default is not None else term.capitalize()
> 
> I would go further and just return `default`. Especially if default
> capitalized term is more a matter of taste as you said. Right now
> it hides the case when the label does not exist. But I'm fine with
> this too.

You're right.  One less selection is almost always a win, and it
was a questionable functionality in the first place.

I was briefly considering whether there's a point in making raising a
KeyError the default, but if people want that, they can directly look
in voc.terms.

pyvo/utils/vocabularies.py
```diff
@@ -0,0 +1,101 @@
...
+IVOA_VOCABULARY_ROOT = "http://www.ivoa.net/rdf/"
+
+
+class VocabularyError(Exception):
+    """A generic error that occurred when interacting with the IVOA
+    vocabulary repository.
+    """
+
+
+if astropy.__version__<'4': # pragma: no cover
```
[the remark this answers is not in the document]
andamian: @msdemlei have you tried to remove this `if` statement?

msdemlei:
On Mon, Mar 22, 2021 at 06:43:39AM -0700, andamian wrote:
> @andamian commented on this pull request.
> +if astropy.__version__<'4': # pragma: no cover
> 
> @msdemlei have you tried to remove this `if` statement?

Can I pull a Bartleby ("I would prefer not to")?

You see, I'm running this on Debian stable, and I would love it if
people on buster could use this without having to upgrade astropy
(which is a bit of a hassle).

However, if it bothers you it can obviously go -- it clearly doesn't
have any effect on current astropys.

andamian:
> On Mon, Mar 22, 2021 at 06:43:39AM -0700, andamian wrote: @andamian commented on this pull request. +if astropy.__version__<'4': # pragma: no cover @msdemlei have you tried to remove this `if` statement?
> Can I pull a Bartleby ("I would prefer not to")? You see, I'm running this on Debian stable, and I would love it if people on buster could use this without having to upgrade astropy (which is a bit of a hassle). However, if it bothers you it can obviously go -- it clearly doesn't have any effect on current astropys.

This would be code that we don't test during the CI but we'll have to support if we release. It's true that we don't have a lower limit on `astropy` dependencies yet which I think we should. If we add that limit, then clients won't be able to install the latest `pyvo` without upgrading `astropy` and this will become dead code.

andamian (approved):

andamian:
@msdemlei thanks for making the changes. I've approved the PR. @funbaker - are you still planing to review it? If not, we'll merge it into master.

Issue #240, opened 2020-09-21
Drop support for Python3.5


andamian:
Both astropy and astroquery are out of 3.5 so I don't see a reason to continue in supporting it.

cbanek:
Sounds good to me!

tomdonaldson:
Agreed!

andamian:
Fixed in #246

PR #242, opened 2020-10-02
Drop 3.5

andamian:
#240


cbanek:
This looks good to me, approving!  Although we might want to squash some of these commits down into a cleaner history?  Maybe 2 commits one for the python and one for the test?  Totally fine keeping them on this branch and PR

cbanek (approved):

.travis.yml
```diff
@@ -76,10 +76,8 @@ matrix:
 
           # try all python versions
         - os: linux
-          env: PYTHON_VERSION=3.6 NUMPY_VERSION=1.15 ASTROPY_VERSION=3.2
+          env: PYTHON_VERSION=3.7 NUMPY_VERSION=1.15 ASTROPY_VERSION=3.2
                PIP_DEPENDENCIES='requests mimeparse requests_mock contextlib2'
-        - os: linux
-          env: PYTHON_VERSION=3.5 NUMPY_VERSION=1.14 ASTROPY_VERSION=2.0.16
```
tomdonaldson, line 82: Should this have been changed to Python 3.6 instead of deleted?  It looks like we wouldn't test 3.6 anymore without it.

tomdonaldson (changes requested):

.travis.yml
andamian: I've dropped it because the job failed:
```This version of Astropy (2.0.16) is incompatible with this version of pytest (6.1.1). Use pytest<3.7.```
I admit I didn't dig further down to try to fix it but I can have a look if you think it's important to keep it.

tomdonaldson: I definitely admit to not knowing all the variables that go into specifying these tests.  Among versions of Python, Astropy and various other support libraries, all the combinations would be silly.  

I was mainly noting that we used to have at least one test at each supported Python version, and now we don't.  It's not a huge concern, but there were some new features added with 3.7, so it would be nice to have a test catch if we inadvertently used one of those features while 3.6 is still supported.  Maybe we could just add a generic test with 3.6 that doesn't care about the versions of the other packages.

andamian: Point taken. Fixed. I've also cleanup `.travis.xml` and brought it more or less in line with what `astroquery` does. Not sure about the `osx` target which doesn't seem to get executed unless I'm missing something.

pllim:
So, did `pyvo` drop Python 3.5 or not? This PR was closed without merge.

pllim:
Ah, nvm. I found #246

PR #246, opened 2020-12-01
Refine GitHub Actions

andamian:
#244

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -3,6 +3,8 @@ name: CI
 on:
   push:
   pull_request:
+  schedule:
+    - cron: "0 3 * * *"
 
 jobs:
   egginfo:
@@ -19,11 +21,10 @@ jobs:
         run: tox -e egg_info
   tests:
     needs: egginfo
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest 
     strategy:
       fail-fast: true
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: [3.7, 3.8, 3.9]
     steps:
     - name: Checkout code
@@ -54,10 +55,40 @@ jobs:
     - name: Python 3.8 with astropy32
       if: matrix.python-version == 3.8
       run: tox -e py38-test-astropy32
+    - name: Python 3.8 with astropy40
+      if: matrix.python-version == 3.8
+      run: tox -e py38-test-astropy40
+    - name: Python 3.8 with astropy41
+      if: matrix.python-version == 3.8
+      run: tox -e py38-test-astropy41
+    - name: Python 3.9 with astropy41
+      if: matrix.python-version == 3.9
+      run: tox -e py39-test-astropy41
     - name: Python 3.9 with latest astropy
       if: matrix.python-version == 3.9
       run: tox -e py39-test
-   
+
+  mac_windows:
+    runs-on: ${{ matrix.os }}
+    needs: tests
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [macos-latest, windows-latest]
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install tox
+      run: python -m pip install --upgrade tox
+    - name: Python 3.8 with latest astropy
+      run: tox -e py38-test
+
   style_docs_coverage:
     runs-on: ubuntu-latest
     needs: tests
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -21,11 +21,13 @@ jobs:
         run: tox -e egg_info
   tests:
     needs: egginfo
-    runs-on: ubuntu-latest 
+    runs-on: ubuntu-latest
+    name: ${{ matrix.python-version }} with ${{ matrix.astropy-version }}
     strategy:
       fail-fast: true
       matrix:
         python-version: [3.7, 3.8, 3.9]
+        astropy-version:  [latest ,astropy40, astropy41]
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
@@ -37,36 +39,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install tox
       run: python -m pip install --upgrade tox
-    - name: Python 3.7 with latest astropy
-      if: matrix.python-version == 3.7
-      run: tox -e py37-test
-    - name: Python 3.7 with astropy30
-      if: matrix.python-version == 3.7
-      run: tox -e py37-test-astropy30
-    - name: Python 3.7 with astropy31
-      if: matrix.python-version == 3.7
-      run: tox -e py37-test-astropy31
-    - name: Python 3.7 with astropy32
-      if: matrix.python-version == 3.7
-      run: tox -e py37-test-astropy32
-    - name: Python 3.8 with latest astropy
-      if: matrix.python-version == 3.8
-      run: tox -e py38-test
-    - name: Python 3.8 with astropy32
-      if: matrix.python-version == 3.8
-      run: tox -e py38-test-astropy32
-    - name: Python 3.8 with astropy40
-      if: matrix.python-version == 3.8
-      run: tox -e py38-test-astropy40
-    - name: Python 3.8 with astropy41
-      if: matrix.python-version == 3.8
-      run: tox -e py38-test-astropy41
-    - name: Python 3.9 with astropy41
-      if: matrix.python-version == 3.9
-      run: tox -e py39-test-astropy41
-    - name: Python 3.9 with latest astropy
-      if: matrix.python-version == 3.9
-      run: tox -e py39-test
+    - name: Python ${{ matrix.python-version }} with astropy
+      run: tox -e py${{ matrix.python-version }}-${{ matrix.astropy-version }}
+
 
   mac_windows:
     runs-on: ${{ matrix.os }}
@@ -87,9 +62,10 @@ jobs:
     - name: Install tox
       run: python -m pip install --upgrade tox
     - name: Python 3.8 with latest astropy
-      run: tox -e py38-test
+      run: tox -e py3.8-latest
 
-  style_docs_coverage:
+
+  style_coverage:
     runs-on: ubuntu-latest
     needs: tests
     steps:
@@ -104,15 +80,13 @@ jobs:
         run: python -m pip install --upgrade tox
       - name: Check codestyle
         run: tox -e codestyle
-      - name: Check docs
-        run: tox -e docs
       - name: Check coverage
         run: tox -e cov
       - name: Upload coverage to codecov
         uses: codecov/codecov-action@v1
         with:
           file: ./coverage.xml
-  
+       
   astropydev:
     # TODO: this job should be allowed to fail except that the functionality
     # is not available in actions yet.
@@ -127,7 +101,7 @@ jobs:
       - name: Install tox
         run: python -m pip install --upgrade tox
       - name: Run tests against astropy dev
-        run: tox -e astropydev
+        run: tox -e py3.8-astropydev
 
   #publish:
   #needs: coverage
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -22,5 +22,3 @@ python:
 
 # Don't build any extra formats
 formats: []
-
-requirements_file: docs/requirements.txt
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,8 +1,26 @@
+version: 2
+
+build:
+  image: latest
+
+sphinx:
+  builder: html
+  configuration: docs/conf.py
+  fail_on_warning: true
+
+# Install regular dependencies.
+# Then, install special pinning for RTD.
 python:
-  version: 3.5
-  pip_install: true
+  version: 3.8
+  system_packages: false
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
+        - all
 
 # Don't build any extra formats
-# formats: []
+formats: []
 
 requirements_file: docs/requirements.txt
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -14,6 +14,7 @@ python:
   version: 3.8
   system_packages: false
   install:
+    - requirements: docs/requirements.txt
     - method: pip
       path: .
       extra_requirements:
--- a/README.rst
+++ b/README.rst
@@ -6,7 +6,7 @@ PyVO
     :alt: Powered by Astropy Badge
 
 .. image:: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
-   :target: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
+    :target: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
     :alt: CI Status
 
 .. image:: https://coveralls.io/repos/github/pyvirtobs/pyvo/badge.svg?branch=master
--- a/README.rst
+++ b/README.rst
@@ -5,8 +5,8 @@ PyVO
     :target: http://www.astropy.org
     :alt: Powered by Astropy Badge
 
-.. image:: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
-    :target: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
+.. image:: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master
+    :target: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master
     :alt: CI Status
 
 .. image:: https://coveralls.io/repos/github/pyvirtobs/pyvo/badge.svg?branch=master
--- a/README.rst
+++ b/README.rst
@@ -5,9 +5,9 @@ PyVO
     :target: http://www.astropy.org
     :alt: Powered by Astropy Badge
 
-.. image:: https://travis-ci.org/astropy/pyvo.svg
-    :target: https://travis-ci.org/astropy/pyvo
-    :alt: Travis Status
+.. image:: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
+   :target: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
+    :alt: CI Status
 
 .. image:: https://coveralls.io/repos/github/pyvirtobs/pyvo/badge.svg?branch=master
     :target: https://coveralls.io/github/pyvirtobs/pyvo?branch=master
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,34 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+# This file is the main file used when running tests with pytest directly,
+# in particular if running e.g. ``pytest docs/``.
+
+import os
+import tempfile
+
+try:
+    from pytest_astropy_header.display import PYTEST_HEADER_MODULES, TESTED_VERSIONS
+except ImportError:
+    PYTEST_HEADER_MODULES = {}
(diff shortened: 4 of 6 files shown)


andamian:
Accomplishments in this PR:
- moved tests to use `pytest` command
- pull out `macOS` and `Windows` out of the matrix into a separate job
- added `astropy40` and `astropy41` to the regression mix
- fixed coverage
- updated the badge in `README`

Outstanding:
- Is the regression mix appropriate? The actual tests that are being executed are listed in the details of the `tests` jobs. Please let me know if you want to change anything in there
- I don't know how to have the coverage jobs show up in the check list similar to `astropy` instead of the bot report. Is that a repo config thing?
- The CI yml file might be a bit to verbose. Wish there was a way to hide the skipped steps in the report. A different option that I was considering was to configure `tox` to ignore tests for which the appropriate Python interpreter was missing. That would make the `yml` file less verbose but the report would be less visible as well (`tox` report would contain all the regression combinations). Anyways, suggestions are welcome.

.github/workflows/cibuild.yml
```diff
@@ -54,10 +55,40 @@ jobs:
     - name: Python 3.8 with astropy32
       if: matrix.python-version == 3.8
       run: tox -e py38-test-astropy32
+    - name: Python 3.8 with astropy40
```
pllim, line 58: I guess this works. `astropy` has a different style if you are interested: https://github.com/astropy/astropy/blob/master/.github/workflows/ci_workflows.yml

.github/workflows/cibuild.yml
```diff
@@ -3,6 +3,8 @@ name: CI
 on:
   push:
```
pllim, line 4: If you want to prevent people from kicking off CI on their own feature branches in their forks, you can make this stricter.
```suggestion
  push:
    branches:
    - master
    tags:
```

README.rst
```diff
@@ -5,9 +5,9 @@ PyVO
     :target: http://www.astropy.org
     :alt: Powered by Astropy Badge
 
-.. image:: https://travis-ci.org/astropy/pyvo.svg
-    :target: https://travis-ci.org/astropy/pyvo
-    :alt: Travis Status
+.. image:: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
+   :target: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
```
pllim, line 9:
```suggestion
    :target: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
```

pllim (commented):
The matrix could be trimmed. Is there a reason why you are testing against astropy 3.x when 4.0.x is the current LTS? I think the important astropy versions to test against are: 4.0.x (LTS), 4.2 (latest stable), and dev

So the matrix could be trimmed down to 3 jobs (any more you add would be a bonus but I am not sure how necessary):

1. Python 3.7 + astropy 4.0.x + your oldest numpy (1.17?)
2. Python 3.8 + astropy 4.2 + latest stable numpy
3. Python 3.9 + astropy dev + latest/dev numpy (dev numpy might be overkill)

The coverage job: Can it be attached to one of the matrix jobs above (maybe stable astropy + stable numpy)? See `astropy` Actions + Tox setup for example.

Coverage reporting: I do see that it commented on the PR, so coverage upload works. `astropy` does have a `codecov.yml` that disables commenting but I don't think that is mutually exclusive with PR status posting. Maybe get this in `master` first and see if the status then magically appears in subsequent PRs.

Doc build job: I would say it is unnecessary if you opt to use RTD PR builder (which also allows doc preview). Up to you though.

.github/workflows/cibuild.yml
andamian: > The matrix could be trimmed. Is there a reason why you are testing against astropy 3.x when 4.0.x is the current LTS? I think the important astropy versions to test against are: 4.0.x (LTS), 4.2 (latest stable), and dev
> 
> So the matrix could be trimmed down to 3 jobs (any more you add would be a bonus but I am not sure how necessary):
> 
> 1. Python 3.7 + astropy 4.0.x + your oldest numpy (1.17?)
> 2. Python 3.8 + astropy 4.2 + latest stable numpy
> 3. Python 3.9 + astropy dev + latest/dev numpy (dev numpy might be overkill)
> 
> The coverage job: Can it be attached to one of the matrix jobs above (maybe stable astropy + stable numpy)? See `astropy` Actions + Tox setup for example.
> 
> Coverage reporting: I do see that it commented on the PR, so coverage upload works. `astropy` does have a `codecov.yml` that disables commenting but I don't think that is mutually exclusive with PR status posting. Maybe get this in `master` first and see if the status then magically appears in subsequent PRs.
> 
> Doc build job: I would say it is unnecessary if you opt to use RTD PR builder (which also allows doc preview). Up to you though.

Thanks @pllim 
I'm fine with dropping support for 3.x if that's `astropy` policy. Do we have agreement. As for `numpy`, as I've mentioned before, for the `numpy` used in this project I don't think we should bother with different versions. That's dealt with upstream (i.e. `astropy`)

I'm going to look into attaching coverage to one of the existing jobs in the matrix

How does the RTD builder app work? I haven't used it before.

.github/workflows/cibuild.yml
andamian: Yes, I will do that if we agree to a regression matrix that is symmetrical.

.github/workflows/cibuild.yml
andamian: Any reason not to do CI on feature branches? I use them in my fork and that's why I prefer to see the CI results after a push in a feature branch.

pllim: > How does the RTD builder app work?

Go to https://readthedocs.org/projects/pyvo/ -> Admin -> Advanced Settings (top left) -> check the "Build pull requests for this project" -> click "Save" (bottom)

Any push to PR after that would trigger a new job from RTD. It is controlled by your https://github.com/astropy/pyvo/blob/master/.readthedocs.yml but you do need to upgrade that YAML to RTD v2 first (https://docs.readthedocs.io/en/stable/config-file/v2.html). You also have the option to let it fail if there is any warning or not; up to you.

> Any reason not to do CI on feature branches?

To save the Earth a little, I guess. But if you think those extra runs are necessary, then keep it as-is.

pllim: As for dropping matrix for `astropy` 3.x, that is up to you. I don't see any minversion of `astropy` defined in your `setup.cfg`. Usually you want to test against the minversion that you say you support.

tomdonaldson: I agree with @pllim's suggestion on trimming the matrix, and see no reason to do anything with `astropy` 3.x.  I would even be OK with us specifying `astropy>=4.1` in `setup.cfg` since that's not a big ask from our users and gives us consistent handling of string values in VOTables (no more byte strings).

For our convenience, I do like triggering CI on feature branches.  It is overkill, but I have found it to save me headaches and dev time by catching weird things sooner.  That helps me rationalize the extra runs along with the fact that the `pyvo` test suite doesn't take long at all to run.

I yield to @pllim's opinion on the other matters, and admit to not fully grasping the details of the macOS and Windows build specs, but am happy to see how they run.

tomdonaldson (changes requested):

.github/workflows/cibuild.yml
andamian: @pllim do you know if `astropy` releases minor versions to work with newer Python version? E.g. recently released P3.9 expected to work with latest `astropy4.0`?

.github/workflows/cibuild.yml
andamian: I guess it's up to all of us to decide on the minversion of `astropy`. Easiest is of course to support just the current version but we should be considerate of users that deploy `pyvo` in pipelines, virtual or containerized environments for which upgrading is not a minor task. Are there any `conda` constraints that we need to consider? How about direct dependencies such as `astroquery`? @bsipocz what is `astroquery` policy on minversions? Otherwise, we could decide to support only minversion that newer then 1 (or 2 or X) years, for example.
In other projects that I'm involved with, users are very slow in upgrading for the reasons above (no sysadmin access, old VM images, etc). I'm still maintaining software that is supposed to run Python2.7 and that becomes trickier! I'm surprised but happy that `astropy` is dropping support for older Python version at this rate.

.github/workflows/cibuild.yml
pllim: astropy 4.0.4 should work with Python 3.9; I see the wheels for Python 3.9 at https://pypi.org/project/astropy/4.0.4/#files

.github/workflows/cibuild.yml
pllim: Is there a way you can ask PyVO stakeholders about this? No one should still be using Python 2 or 3.5, so I don't see why anyone needs `astropy<4`.

andamian: I think that `astroquery` is the main stakeholder so I'll wait for @bsipocz word.

pllim: Can ping @keflavich and @ceb8 too.

keflavich: I don't think we rely on astropy < 4; is that the question?  haven't time to read back through whole convo now

andamian: Yes, that was the question. In this case, we'll drop support for 3.*

andamian:
Another version with trimmed matrix is now available.

.github/workflows/cibuild.yml
```diff
@@ -3,6 +3,8 @@ name: CI
...
   pull_request:
+  schedule:
```
tomdonaldson, line 6: I've just noticed this section.  What is the intended effect of this `cron`?  I'm new to this syntax so it looks like this scheduling is independent of the other triggers (so the CI would run every 3 hours all the time), but I could also imagine it somehow chaining with the push and pull_request to bin the CI runs once those triggers happen.  Are there reasons to run other than when a push or pull_request trigger happens?

pllim: I think that spec tells it to run **daily** at 3 AM (UTC). If this is a low-traffic package, weekly might work too.

Cron job is to catch upstream changes that might break your code. So I don't think you should get rid of it. Most of the time, it would pass, but for that rare case a new upstream release breaks your code, you would want to know even when there is no push to `master`. Most common example is that a new release finally removed that deprecated API and you have been ignoring the deprecation warning; or it removed/changed a hidden function that you have been using. Hope that clarifies the matter.

.github/workflows/cibuild.yml
```diff
@@ -36,29 +39,33 @@ jobs:
...
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install tox
+      run: python -m pip install --upgrade tox
     - name: Python 3.8 with latest astropy
```
pllim, line 64: Why latest and not stable `astropy` here for OSX and Windows?

pllim: Ah, nvm... I confused myself. Ignore the comment above.

.readthedocs.yml
```diff
@@ -1,8 +1,26 @@
...
+      path: .
+      extra_requirements:
+        - docs
+        - all
 
 # Don't build any extra formats
-# formats: []
+formats: []
 
 requirements_file: docs/requirements.txt
```
pllim, line 26: Don't need this line anymore?
```suggestion
```

README.rst
```diff
@@ -5,9 +5,9 @@ PyVO
     :target: http://www.astropy.org
     :alt: Powered by Astropy Badge
 
-.. image:: https://travis-ci.org/astropy/pyvo.svg
-    :target: https://travis-ci.org/astropy/pyvo
-    :alt: Travis Status
+.. image:: https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master&event=schedule
```
pllim, line 8: Why `event=schedule`?

tox.ini
```diff
@@ -12,13 +11,13 @@ requires =
...
     pip freeze
-    python setup.py test
+    pytest --pyargs pyvo
 
 description =
     run tests
-    astropy30: with astropy 3.0.*
-    astropy31: with numpy 3.1.*
-    astropy32: with numpy 3.2.*
+    astropy40: with astropy 4.0.*
```
pllim, line 18: You might also want to match your `astropy` minversion to your `numpy` minversion for cross-testing.

pllim:
p.s. Weird, I don't see the RTD build here. Did you enable it on the RTD side?

pllim:
p.p.s. Is `egginfo` test still necessary?

pllim:
p.p.p.s.

Test header won't show unless you move https://github.com/astropy/pyvo/blob/master/pyvo/conftest.py to root directory. Some people say that messes up the test runner (i.e., `mypackage.test()`) but not seeing it in CI also isn't very useful.

Also, I was thinking more like 3 jobs instead of 9 jobs.

* Python 3.7 + astropy 4.0
* Python 3.8 + astropy 4.1
* Python 3.9 + astropy 4.2 (latest)

.github/workflows/cibuild.yml
andamian: The nightly build cron runs at 3:00UTC in the master.

README.rst
andamian: Because it reflects the (more or less current) status of the `master` branch. Where should we point it to?

pllim: Maybe https://github.com/astropy/pyvo/workflows/CI/badge.svg?branch=master is enough so it picks up merge commit runs too?

tox.ini
andamian: Sorry, I don't understand. We don't have min versions specified and we've decided not to do cross-testing here, no?

pllim: In that case, I still recommend cross testing with the minversion of numpy that astropy 4.0 supports (numpy 1.16), but if you disagree, feel free to ignore; this is just a recommendation.

andamian:
> p.s. Weird, I don't see the RTD build here. Did you enable it on the RTD side?

I thought I did.

andamian:
> p.p.s. Is `egginfo` test still necessary?

It was there from before so I just preserved it. I assume it was meant to do a quick check before spawning all the test jobs.

pllim:
Re: egginfo -- Ah, okay.

Re: RTD -- Ops, I forgot. The webhook setting in this repo also has to be updated. I have updated it for you. So theoretically, in your next push, it should kick off. 🤞

andamian:
> p.p.p.s.
> 
> Test header won't show unless you move https://github.com/astropy/pyvo/blob/master/pyvo/conftest.py to root directory. Some people say that messes up the test runner (i.e., `mypackage.test()`) but not seeing it in CI also isn't very useful.
> 

What is the test header?

> Also, I was thinking more like 3 jobs instead of 9 jobs.
> 
> * Python 3.7 + astropy 4.0
> * Python 3.8 + astropy 4.1
> * Python 3.9 + astropy 4.2 (latest)

Really? 3 doesn't make a matrix :-). 
Do you think there are too many jobs or some combinations are not meaningful? I can see for example that Py3.9 + Astropy 4.0 might not be that relevant since a user that just upgraded to Py3.9 is likely going to get the latest version of the libraries but that is not always the case. And it's perfectly reasonable for Py3.8 users to want to upgrade to latest `astropy`, no?
If the problem is the number of jobs, we could do the regression with `tox` so that there will be 3 jobs but each of them will test with 3 versions of `astropy`. It will be less clear what failed and why. What do you think?

andamian:
> Re: egginfo -- Ah, okay.
> 
> Re: RTD -- Ops, I forgot. The webhook setting in this repo also has to be updated. I have updated it for you. So theoretically, in your next push, it should kick off. 🤞

Awesome! Thanks!

pllim:
> What is the test header?

https://github.com/astropy/pytest-astropy-header ; it reports the dependency versions. I see that you have it set up. It is just not displaying in the CI log.

> Do you think there are too many jobs or some combinations are not meaningful?

It all depends what you want to test, I guess. Is it really PyVO's problem when astropy 4.0 suddenly breaks in Python 3.9? Do you need to test all those combos for every commit on a PR?

I can see that it might be useful for cron, but might be a little excessive for pull request? But if you disagree, feel free to keep it as is. I'll leave that to your good judgement.

.github/workflows/cibuild.yml
tomdonaldson: Yes, sorry, I don't know why I read that as every 3 hours.  Daily seems OK to me.  It's nice to see those upstream impacts sooner rather than later.

andamian:
> > What is the test header?
> 
> https://github.com/astropy/pytest-astropy-header ; it reports the dependency versions. I see that you have it set up. It is just not displaying in the CI log.

Got it. I've moved it to `root` - hopefully this helps.
> 

> It all depends what you want to test, I guess. Is it really PyVO's problem when astropy 4.0 suddenly breaks in Python 3.9? Do you need to test all those combos for every commit on a PR?
> 
> I can see that it might be useful for cron, but might be a little excessive for pull request? But if you disagree, feel free to keep it as is. I'll leave that to your good judgement.

I agree with you that it might be excessive but `pyvo` is relatively lightweight and tests run fast. For now, I would keep it in this form unless the others have a different opinion. Later we can revisit it and trim it down if necessary or split the flow into a `ci` and a `nightly` for ex. Good point nevertheless. Thanks @pllim

andamian:
@pllim - good news is that RTD kicked, the bad news is that it failed. Do you happen to know why `sphinx_astropy` does not get installed? Where should I add it as a dependency?

.readthedocs.yml
```diff
@@ -1,8 +1,24 @@
...
-  version: 3.5
-  pip_install: true
+  version: 3.8
+  system_packages: false
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs
+        - all
```
pllim, line 21: I see that you have not switched to declaring your dependencies in `setup.cfg` yet, so maybe try this?

```suggestion
    - requirements: docs/requirements.txt
```

conftest.py
```diff
@@ -0,0 +1,26 @@
...
+# This file is the main file used when running tests with pytest directly,
+# in particular if running e.g. ``pytest docs/``.
+
+import os
+import tempfile
+
+try:
+    from pytest_astropy_header.display import PYTEST_HEADER_MODULES
+except ImportError:
+    PYTEST_HEADER_MODULES = {}
```
pllim, line 12: Need this
```suggestion
try:
    from pytest_astropy_header.display import PYTEST_HEADER_MODULES, TESTED_VERSIONS
except ImportError:
    PYTEST_HEADER_MODULES = {}
    TESTED_VERSIONS = {}

try:
    from pyvo import __version__ as version
except ImportError:
    version = 'unknown'

TESTED_VERSIONS['pyvo'] = version
```

and also this in `setup.cfg` under this section (I think your `[tools:pytest]` is a typo):

```
[tool:pytest]
astropy_header = true
```

andamian:
@tomdonaldson, @cbanek & @bsipocz - This is ready for the final review. @pllim has helped me iron out the last details (many thanks!). Please review at your earliest convenience or let me know if you are OK with me releasing it. There is at least another PR that is blocked on the CI and I'd like to get back to the user soon. Thanks.

tomdonaldson (approved):

tomdonaldson:
> @tomdonaldson, @cbanek & @bsipocz - This is ready for the final review. @pllim has helped me iron out the last details (many thanks!). Please review at your earliest convenience or let me know if you are OK with me releasing it. There is at least another PR that is blocked on the CI and I'd like to get back to the user soon. Thanks.

This all looks fine to me.  Thanks very much @pllim for the discussion and guidance.  I've learned a lot.

.github/workflows/cibuild.yml
cbanek: Yeah, I'm also nervous about build breakages due to outside things, even outside astropy, as sometimes it's a while between PRs.

cbanek (approved):
Looks good to me, so many thanks to @andamian for doing all the work and @pllim for all the great feedback.  I echo @tomdonaldson that I learned a lot!

pllim (approved):
The test header is still not appearing but it can be fixed as a separate PR, because I think you might as well move the dependencies metadata to `setup.cfg` and apply APE 17 while you are at it.

Thanks for your hard work on this!

p.s. You are welcome! Glad to help.

Issue #238, opened 2020-09-17
pyvo.registry search etc: option to show auxiliary capabilities


theresadower:
Currently, the pyvo.registry search and ivoid2service functions match on servicetypes in a way that excludes auxiliary capabilities, like TAP endpoints associated with a simpler ConeSearch record.  While this is fine in many cases and may be an explicit design decision, we've found in working with NAVO tutorial notebooks that it could be very handy and in some cases an easier search pattern if the user could see these.

Requesting feature flag to include auxiliary capability information or add it to the default behavior. I'm inclined to add a flag and happy to write it myself either way, but wanted input on preference from more seasoned pyvo users/maintainers, or to be pointed at documentation if I've missed a way for this to be enumerated already.

Cheers!

msdemlei:
On Thu, Sep 17, 2020 at 01:38:28PM -0700, theresa dower wrote:
> Requesting feature flag to include auxiliary capability information
> or add it to the default behavior. I'm inclined to add a flag and
> happy to write it myself either way, but wanted input on preference
> from more seasoned pyvo users/maintainers, or to be pointed at
> documentation if I've missed a way for this to be enumerated
> already.

I believe for pyVO is most common discovery case is "all-VO query
with constraints", which is service enumeration in the language of
http://ivoa.net/documents/discovercollections/: Essentially, you want
to hit every service of a given type exactly once.

Hence, I'd rather confidently say the default should remain the
exclusion of auxiliary capabilities.

Supporting data discovery (and thus aux capabilities), of course, is
still a good idea, in particular when we expect usage through
notebook-like interfaces.  Out of curiosity: Do you have a particular
scenario in mind?

theresadower:
Agreed that the default behavior should skip them, especially if we're in most use cases looping over results at the resource level. I'll work on a feature with a flag for including aux and not touch the defaults.

The scenario in which this came up was an evolution of a use case notebook in https://github.com/NASA-NAVO/navo-workshop: we are looking for catalog data from a specific mission, finding the cone search, and then doing a deeper dive into more complex TAP queries. The ability to follow logical links from the HEASARC's ConeSearch service for the data as found in the registry to the main HEASARC xamin TAP service hit a snag when the aux capability doesn't come back.

msdemlei:
Uh, apologies I let this drop in September.

[...]

PR #258, opened 2021-04-05
RegTAP search: add a flag for including standard auxiliary capabilities

theresadower:
Proposed search flag includeaux : boolean, default false. 

This flag controls whether to include auxiliary capabilities (with standard interfaces) in registry search results. This may result in duplicate capabilities being returned, especially if the servicetype is not specified, and therefore remains default off.

This behavior is discussed in https://github.com/astropy/pyvo/issues/238 . The issue originally arose in a NAVO-run pyvo workshop at AAS, where a workflow of performing cone searches and then moving to a related TAP service for more fine-grained queries and additional filtering didn't work as expected. @trjaffe and @tomdonaldson can comment further from the workshop experience.


cbanek:
I'm not a registry expert, but this looks pretty good.  I'm not sure what's up with the test failures, it looks like maybe those are related to some kind of mock issue, but once the tests pass I'm good to approve!

Thanks for the PR!

theresadower:
> 
> 
> I'm not a registry expert, but this looks pretty good. I'm not sure what's up with the test failures, it looks like maybe those are related to some kind of mock issue, but once the tests pass I'm good to approve!
> 
> Thanks for the PR!

Thanks! I'm not very familiar with fixtures or mocker myself. I'll spend some time working on this and possibly ask around for help, then un-mark it WIP for further review once they pass.

theresadower:
@cbanek  Tests fixed. It looks like one of the remaining failures is needing a milestone; I don't understand the codestyle: commands failed error, caused by the former or otherwise. Please let me know what I should do next, if anything, and thank you for your time!

pyvo/registry/tests/test_regtap.py
```diff
@@ -138,6 +138,22 @@ def datamodeltest_callback(request, content):
     ) as matcher:
         yield matcher
 
+@pytest.fixture()
```
cbanek, line 141: I think it's looking for one more empty line above here

cbanek (approved):
LGTM.  Milestone set.  Leaving for anyone else open to comment for a while.  While I think the where clauses that are strings could be maybe combined or something, it's probably better to not complicate that part, since it's really readable the way it is here (and there's all sorts of issues with parens and other things like that).

msdemlei:
As commented over in issue #238, this isn't my preferred way to address the underlying use case.  

But if we do it this way, we shouldn't blindly match anything that ends in aux; that will give far too many false positives (try ``servicetype='sia' includeaux=True``).  If we go the way proposed here, I'd say we ought to have something like

```
# where perhaps keep this in a central location
KNOWN_SERVICE_TYPES = {"conesearch", "sia", "ssa", "slap", "tap"}

match_caps = None  # not used with this logic, but for robustness
if servicetype:
  match_caps = set([servicetype])
else:
  match_caps = KNOWN_SERVICE_TYPES

if includeaux:
  match_caps |= {s+"#aux" for s in match_caps}

if match_caps:
  wheres.append("standard_id IN ({})".format(
    ",".join(
      tap.as_adql_literal("ivo://ivoa.net/std/"+s)
      for s in match_caps)))
```

``tap.as_adql_literal`` would have to be defined (but that's a useful function anyway) -- it would return ADQL literals for python values, e.g., ``'that''s'`` for a python string ``that's``.

(I'd like that better than current main's code whether or not we merge this PR)

msdemlei:
Incidentally: Whatever we do, I suppose we need to touch the service property, too, when we deal with auxiliary capabilities.  For now, a startwith approach as in describe would look about ok, but I have to admit I've not really looked at this particular part of the API and frankly suspect this needs to be re-thought a bit because of course there are now many resources that are both, say, SCS and TAP.

theresadower:
> 
> 
> As commented over in issue #238, this isn't my preferred way to address the underlying use case.
> 
> But if we do it this way, we shouldn't blindly match anything that ends in aux; that will give far too many false positives (try `servicetype='sia' includeaux=True`). If we go the way proposed here, I'd say we ought to have something like
> 
> ```
> # where perhaps keep this in a central location
> KNOWN_SERVICE_TYPES = {"conesearch", "sia", "ssa", "slap", "tap"}
> 
> match_caps = None  # not used with this logic, but for robustness
> if servicetype:
>   match_caps = set([servicetype])
> else:
>   match_caps = KNOWN_SERVICE_TYPES
> 
> if includeaux:
>   match_caps |= {s+"#aux" for s in match_caps}
> 
> if match_caps:
>   wheres.append("standard_id IN ({})".format(
>     ",".join(
>       tap.as_adql_literal("ivo://ivoa.net/std/"+s)
>       for s in match_caps)))
> ```
> 
> `tap.as_adql_literal` would have to be defined (but that's a useful function anyway) -- it would return ADQL literals for python values, e.g., `'that''s'` for a python string `that's`.
> 
> (I'd like that better than current main's code whether or not we merge this PR)

OK, double checking just from a read of the code without having run it, the intention is:
if aux is included and servicetype is specified, only return auxiliary capabilities of that servicetype (that also match the rest of keyword WHEREs, etc) and if aux is included but no servicetype, return all standard capabilities and aux capabilities (that also match the rest of the keyword WHEREs, etc) and (while they would be arguably wrong from the registry record creation and organization standpoint) non-standard aux capabilities, if they ever did exist, would not be returned.

I think I like this actually, @trjaffe, is this closer to what we wanted out of the workshop workflow?

theresadower:
> 
> 
> As commented over in issue #238, this isn't my preferred way to address the underlying use case.
> 
> But if we do it this way, we shouldn't blindly match anything that ends in aux; that will give far too many false positives (try `servicetype='sia' includeaux=True`). If we go the way proposed here, I'd say we ought to have something like
> 
> ```
> # where perhaps keep this in a central location
> KNOWN_SERVICE_TYPES = {"conesearch", "sia", "ssa", "slap", "tap"}
> 
> match_caps = None  # not used with this logic, but for robustness
> if servicetype:
>   match_caps = set([servicetype])
> else:
>   match_caps = KNOWN_SERVICE_TYPES
> 
> if includeaux:
>   match_caps |= {s+"#aux" for s in match_caps}
> 
> if match_caps:
>   wheres.append("standard_id IN ({})".format(
>     ",".join(
>       tap.as_adql_literal("ivo://ivoa.net/std/"+s)
>       for s in match_caps)))
> ```
> 
> `tap.as_adql_literal` would have to be defined (but that's a useful function anyway) -- it would return ADQL literals for python values, e.g., `'that''s'` for a python string `that's`.
> 
> (I'd like that better than current main's code whether or not we merge this PR)

Adapted this slightly - we already had a map of accepted servicetypes to standards-names, where users input 'scs' rather than 'conesearch', and used that rather than break it. Given the additional string wrangling around that, the new adql_as_literal does not seem to be needed.

If we want to change around this behaviour on a larger scale we can do a separate branch and MR for it.

Additionally, added a check for bad servicetype input, added to pytest.

theresadower:
@cbanek This is currently failing due to lack of changelog, forgive me for not knowing process or being able to find it, should I just edit CHANGES.rst or is there some other process that will pick it up?

andamian:
> @cbanek This is currently failing due to lack of changelog, forgive me for not knowing process or being able to find it, should I just edit CHANGES.rst or is there some other process that will pick it up?

@theresadower - you need to update the `CHANGES.rst` and push it. Thanks

cbanek:
This looks good to me, is everyone happy with this and ready to go?

tomdonaldson:
> This looks good to me, is everyone happy with this and ready to go?

Looks good to me too.

Issue #259, opened 2021-04-08
tox python targets are incorrect


andamian:
Default tox targets for python are `py3[7,9]` (not dot) so the current ones are incorrect. CI works only because it installs the correct version of Python for each run but when run locally tox targets are wrong and use the default python instead of a specified version.

PR #260, opened 2021-04-08
Fixed wrong tox targets

andamian:
#259

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -40,7 +40,11 @@ jobs:
     - name: Install tox
       run: python -m pip install --upgrade tox
     - name: Python ${{ matrix.python-version }} with astropy
-      run: tox -e py${{ matrix.python-version }}-${{ matrix.astropy-version }}
+      run: |
+        # remove "." from matrix.python-version to get the defaults tox python versions
+        tox_python_target=$(echo py${{matrix.python-version}} | sed 's/\.//')
+        echo "Tox python target: " $tox_python_target
+        tox -e $tox_python_target-${{ matrix.astropy-version }}
 
 
   mac_windows:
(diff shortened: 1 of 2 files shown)


andamian:
@pllim or @bsipocz - do you have a minute to take a quick look at this? It's quite simple and straightforward but if you have a better idea please let me know. Thanks

.github/workflows/cibuild.yml
```diff
@@ -40,7 +40,11 @@ jobs:
     - name: Install tox
       run: python -m pip install --upgrade tox
     - name: Python ${{ matrix.python-version }} with astropy
-      run: tox -e py${{ matrix.python-version }}-${{ matrix.astropy-version }}
+      run: |
+        # remove "." from matrix.python-version to get the defaults tox python versions
+        tox_python_target=$(echo py${{matrix.python-version}} | sed 's/\.//')
```
pllim, line 45: I think it is less confusing to set `tox_python_target` as another parameter for the `matrix` above, rather then doing some `sed` calls. This won't work if you want to inject other OS in the future.

andamian: Not sure I understand what you mean. Add another dimension to the matrix and only run if `python-version` matches the `tox_python_target`?

andamian: There is some package that enables explicit mapping of GH python version with the `tox` target but it doesn't seem worth it.

pllim: I was thinking something like https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations but now that I look at it again, I realize maybe that's not what you want.

If this syntax works for you for now, I guess it is okay to deploy as is.

andamian: It does the trick for now. If it breaks with future version then will address it at that time I guess. Thanks for your quick feedback!

cbanek (approved):

PR #265, opened 2021-06-05
Fix for #228 that sometimes fails installs on fresh Python environments.

jpl-jengelke:
Add wheel to the `install_requires` setup config for the time being.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,7 @@ license = BSD
 url = https://github.com/astropy/pyvo
 edit_on_github = False
 github_project = astropy/pyvo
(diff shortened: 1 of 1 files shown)


andamian:
@jpl-jengelke - thanks for the nice investigation. If wheels get published with every `PyVO` release this should not be an issue anymore. Agree?

jpl-jengelke:
> @jpl-jengelke - thanks for the nice investigation. If wheels get published with every `PyVO` release this should not be an issue anymore. Agree?

Yes, absolutely. The PR may be closed without merge if you like.

Issue #272, opened 2021-09-21
UWS parser doesn't know errorSummary?


msdemlei:
The following programme:

```
from pyvo.dal import tap

svc = tap.TAPService("http://dc.g-vo.org/tap")
with svc.submit_job("Doofus") as job:
9job.run()
9job.wait()
```

yields:

```
WARNING: UnknownElementWarning: None:1:995: UnknownElementWarning: Unknown element errorSummary [pyvo.utils.xml.elements]
WARNING: UnknownElementWarning: None:1:1044: UnknownElementWarning: Unknown element message [pyvo.utils.xml.elements]
```

although the document returned is UWS- and XSD-valid.  Does anyone
remember why that is so?

If someone is currently working with the code: would you add support for
errorSummary?  Otherwise, please poke me and I'll fix it.

pllim:
I think this was fixed in #276.

Issue #270, opened 2021-08-16
simple_service_discovery.ipynb notebook gives warnings in cell#2 (related to Issue#257 ?)


KenMighell:
```pyvo.__version__='1.1'``` 

The Jupyter notebook  ```pyvo/examples/notebooks/simple_service_discovery.ipynb```  issues warnings in cell#2. 

This problem _may_ be related to Issue #257 .

This script

```
#!/usr/bin/env python3
import pyvo
services = pyvo.regsearch(servicetype="sia", waveband="x-ray")
# EOF
```
produces these warnings

```
WARNING: UnknownElementWarning: None:3:1106: UnknownElementWarning: Unknown element mirrorURL [pyvo.utils.xml.elements]
WARNING: UnknownElementWarning: None:125:4415: UnknownElementWarning: Unknown element mirrorURL [pyvo.utils.xml.elements]
WARNING: UnknownElementWarning: None:125:4732: UnknownElementWarning: Unknown element mirrorURL [pyvo.utils.xml.elements]
WARNING: UnknownElementWarning: None:125:5044: UnknownElementWarning: Unknown element mirrorURL [pyvo.utils.xml.elements]
WARNING: UnknownElementWarning: None:125:5344: UnknownElementWarning: Unknown element mirrorURL [pyvo.utils.xml.elements]
```

Inserting these two lines after importing pyvo stops the warnings from being displayed

```
import warnings
warnings.filterwarnings("ignore", module="pyvo.utils.xml.*")
```

but at the cost of losing all XML parsing diagnostics.

**Comment by KenMighell** [comment]
===== More notebook problems =====

(1) The other Jupyter notebook ```pyvo/examples/notebooks/working_with_registry_results.ipynb``` gives similar warnings (```Unknown element mirrorURL```) in three different cells: cell#1, cell#5, and cell#7.

(2) The code in cell#6:
```
nvss = vo.registry.ivoid2service('ivo://nasa.heasarc/skyview/nvss',servicetype='sia')
nvss.search(pos=(350.85, 58.815),size=0.25,format="image/fits").to_table()
```
gives the following error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-892a8a3f83d6> in <module>
----> 1 nvss = vo.registry.ivoid2service('ivo://nasa.heasarc/skyview/nvss',servicetype='sia')
      2 nvss.search(pos=(350.85, 58.815),size=0.25,format="image/fits").to_table()

[...]

Issue #267, opened 2021-07-02
pyvo installs mimeparse 0.1.3 although python-mimeparse 1.6.0 is already installed


gcomes:
When I install pyvo with pip, it downloads mimeparse 0.1.3, but on my machine I already have python-mimeparse 1.6.0.
Unless there is a requirement to use such ancient version of mimeparse I think the problem is due to the fact that in the file setup.cfg the line:

 install_requires = astropy requests mimeparse

should be changed to:

 install_requires = astropy requests python-mimeparse

pllim:
I think this was fixed in #276 .

PR #276, opened 2021-10-29
Patch for a few issues

andamian:
#272 - @msdemlei - I'm not 100% sure the fix is correct but I couldn't figure out how to make  `message` part of the `errorSummary` class. If you can suggest something better I'll do it.

#270 - I just had to remove the `decode()`. I assume those where there from where `VOTable` was returning the wrong type.

#267 - I've checked and `python-mimeparse` was also in limbo for a while. The functionality we were using from that package was minimal so I've changed the code to not depend on it anymore - pretty small change.


andamian:
@msdemlei - thanks for the quick review
(a) - good catch. I've added tests for `mime type` (hopefully didn't go overboard with it :-)). The error is now a generic `ValueError` - Don't think a special error would add any value.
(b) - removed the result entry from the failed job
(c) - not sure what you mean here. I couldn't find any reference for `errorSummary` previous to my changes.

msdemlei:
On Fri, Oct 29, 2021 at 11:50:44AM -0700, Adrian wrote:
> (c) - not sure what you mean here. I couldn't find any reference for `errorSummary` previous to my changes.

I was just talking about the code coverage; you see, according to my
review interface, for instance the setter for type on ErrorSummary
(currently on line 417) is never executed by tests.  That seems odd
to me, since the example document does have a type attribute.

If you understand why that is, I'd not object too much (I'm no 100%
coverage zaelot and can live with the occasional untested code path),
but if it's not clear why that doesn't get exercised, then I'd
suspect a bug.

andamian:
> On Fri, Oct 29, 2021 at 11:50:44AM -0700, Adrian wrote: (c) - not sure what you mean here. I couldn't find any reference for `errorSummary` previous to my changes.
> I was just talking about the code coverage; you see, according to my review interface, for instance the setter for type on ErrorSummary (currently on line 417) is never executed by tests. That seems odd to me, since the example document does have a type attribute. If you understand why that is, I'd not object too much (I'm no 100% coverage zaelot and can live with the occasional untested code path), but if it's not clear why that doesn't get exercised, then I'd suspect a bug.

Fixed this in `ErrorSummary` and a few other places - please double-check. I've updated other classes so setting a value goes through the property "setter" so that the value could be properly validated. Let me know you agree or you don't.

msdemlei (approved):
Looks good to me, except I can't see why you're adding pillow to the dependencies -- is that an accident?

andamian:
> Looks good to me, except I can't see why you're adding pillow to the dependencies -- is that an accident?

It's now testing https://github.com/andamian/pyvo/blob/3c8de6866fd55b587cc03a4e92de4ae679a032b3/pyvo/dal/mimetype.py#L95

I guess the dependency could go into `pip-requirements` except that I see that file being commented out in `tox.ini` for reasons that I don't recall right now.

tomdonaldson (approved):
Looks good to me.

PR #284, opened 2022-01-07
Add PyPI publish workflow

### LINKED ISSUE #181: Simplify the release process

### LINKED ISSUE #247: Publish to PyPI using GitHub Actions

pllim:
- [x] Needs someone with admin access at https://pypi.org/project/pyvo/ to generate and add token as repo secret (`pypi_password`).
- [x] Devs need to confirm if universal wheel is sufficient.

With this workflow, the new release procedure would be something like this, though you need to adapt it to your existing release procedure (e.g., if you do branching and all that):

1. Finalize change log.
2. Make sure the commit you will tag on does not have `[ci skip]` or `[skip ci]` in its message. (If you do, the workflow won't run.)
3. Tag `main` (customize this instruction and workflow YAML if you do not release from `main`) with release tag: `git tag -s "vX.Y.Z" -m "Tagging version vX.Y.Z"` (you need GPG for `-s` to work, so if you don't do GPG, also customize this command)
4. Push the tag and check "Release" action. If green, proceed. If not, fix the problem and re-tag (and force push it). Make sure it gets on PyPI.
5. (This part needs [Zenodo webhook to listen to GitHub Release](https://docs.github.com/en/repositories/archiving-a-github-repository/referencing-and-citing-content). I see you already have DOI, so I am not sure if you can keep the existing DOI unless you already used that to make the webhook already, you need to check with Zenodo.) Make a GitHub Release off the tag and publish it. Check Zenodo DOI to make sure it got pushed properly.
6. Prepare change log for next release.
7. Move open issues/PRs to next milestone. Close the release milestone.

Fix #247 , fix #181

p.s. Ops... I just saw #266 after opening this...

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -32,12 +32,13 @@ jobs:
     - name: Check long_description
       run: python -m twine check --strict dist/*
 
-    - name: Test package
-      run: |
-        cd ..
-        python -m venv testenv
-        testenv/bin/pip install pytest pytest-astropy mimeparse requests_mock pillow pyvo/dist/*.whl
-        testenv/bin/pytest pyvo
+    # FIXME: Failed due to astropy-helpers doing things.
+    #- name: Test package
+    #  run: |
+    #    cd ..
+    #    python -m venv testenv
+    #    testenv/bin/pip install pytest pytest-astropy mimeparse requests_mock pillow pyvo/dist/*.whl
+    #    testenv/bin/pytest pyvo
 
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,47 @@
+name: Release
+
+on:
+  pull_request:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,48 @@
...
+    # FIXME: Failed due to astropy-helpers doing things.
+    #- name: Test package
+    #  run: |
+    #    cd ..
+    #    python -m venv testenv
+    #    testenv/bin/pip install pytest pytest-astropy mimeparse requests_mock pillow pyvo/dist/*.whl
+    #    testenv/bin/pytest pyvo
+
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags')
```
andamian, line 44: Shouldn't check here if the tag version matches the versions in `setup.cfg` like it does in #266?

andamian:
Thanks for reviving this @pllim.

I thing that (3) can be easily done from the browser with GH Releases.

No experience with Zenodo but I can look into it if needed. However I'm not admin to this repo so I don't know if I can set it up.

.github/workflows/publish.yml
pllim: Oh, I overlooked the fact that you still uses the old infrastructure without `setuptools_scm`. Ops! Maybe this PR is not ideal then unless #285 can be done first.

pllim:
> I'm not admin to this repo so I don't know if I can set up the Webhooks without that

I am not sure either. I do think an admin has to do this, so we should ping the admins of this repo.

pllim:
As discussed today, I will close this PR and review/suggest over at #266 instead.

PR #299, opened 2022-02-09
Handle custom terms in datalink.bysemantics again.

### LINKED ISSUE #298: support for full-url semantics no longer provided, breaking existing code
At ESO we publicised the usage of pyvo, and provided jupyter notebooks to our community showing examples on how to programmatically  access the ESO archive. 
One important feature is the ability to address a particular record in a datalink response using its semantics, e.g.:
`
semantics = 'http://archive.eso.org/rdf/datalink/eso#calSelector_raw2master'
raw2master_url = next(datalink.bysemantics( semantics )).access_url
`

That code worked nicely with pyvo v1.1; with a later version of pyvo (we tried with v1.2.1) it does no longer work.
 
The comment in the current github code (adhoc.py) reads:
`semantics: str or list
            One or more term(s) from the datalink vocabulary
            (http://www.ivoa.net/rdf/datalink/core).  Terms may be
            passed in with or without a leading hash.  Note that at
            this point non-datalink/core URIs (i.e., full resource
            URIs) are not supported (and will be butchered).`

That comment, not present in pyvo v1.1, explains why the ESO code no longer works, as 
the semantics above is a full resource URI (it cannot be changed to a term from the core vocabulary).

So, at this point our beloved pyvo module is no longer compliant with the datalink ivoa standard, as the standard still foresees non-core, full resource URIs. 

Even worse: any script that our ESO user community would have developed in the meantime would break.
There are some instrument operation teams using that functionality. So the impact of this is not minor on us.

May I please ask you to provide a workaround?
Can I also please ask you to re-instantiate the support for full URIs, so to be compliant to the standard?

Many thanks in advance!

msdemlei:
This is done by extracting non-core terms -- really, anything with a URI
that starts with something that is not the core URI -- and excepting
it from our normalisation.

This would fix #298.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 1.3 (unreleased)
 ================
 
+- pyvo deals with non-core terms in datalink.bysemantics again. [#299]
 
 1.2.1 (2022-01-12)
 ==================
--- a/pyvo/dal/adhoc.py
+++ b/pyvo/dal/adhoc.py
@@ -525,10 +525,10 @@ class DatalinkResults(DatalinkResultsMixin, DALResults):
         ----------
         semantics: str or list
             One or more term(s) from the datalink vocabulary
-            (http://www.ivoa.net/rdf/datalink/core).  Terms may be
-            passed in with or without a leading hash.  Note that at
-            this point non-datalink/core URIs (i.e., full resource
-            URIs) are not supported (and will be butchered).
+            (http://www.ivoa.net/rdf/datalink/core).  datalink/core
+            terms may be passed in with or without a leading hash.
+            Free URIs may also be passed an and will be compared literally,
+            i.e., without any URI normalisation.
         include_narrower: boolean
             If true, the result will include matches for any term
             that is narrower than the term passed in.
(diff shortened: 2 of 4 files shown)


msdemlei:
On Thu, Feb 10, 2022 at 05:52:46PM -0800, Adrian wrote:
> > @@ -1,6 +1,8 @@
>  1.3 (unreleased)
>  ================
>  
> +- pyvo deals with non-core terms in datalink.bysemantics again. 
> +  This would fix issue #298.
> 
> You need to append the PR number (`[#299]`) to make the validation

Put that in, forced-pushed.

> happy. If this is an issue that requires quick release, we should
> do a minor release - 1.2.2. @tomdonaldson - do you agree?

A minor release would be lovely, as it'd put lift quite a load from
my consciousness...

> >          if isinstance(semantics, str):
>              semantics = [semantics]
> -        semantics = [term.lstrip("#") for term in semantics]
> +
> +        core_terms, other_terms = [], []
> 
> Is there a reason to use lists instead of sets?

No -- one *could* work with sets all along and not just at the end.
There's not much benefit to performance or clarity either way,
though, so unless there's an aspect to this I'm missing I'd leave it
this way.

msdemlei:
I've asked the bug reporter about a verifier, and he offered

{{{
import pyvo
datalink_url = 'http://archive.eso.org/datalink/links?ID=ivo://eso.org/ID?SPHER.2016-09-26T03:04:09.308'
datalink = pyvo.dal.adhoc.DatalinkResults.from_result_url(datalink_url)
semantics = 'http://archive.eso.org/rdf/datalink/eso#calSelector_raw2master'
raw2master_url = next(datalink.bysemantics( semantics )).access_url
print(raw2master_url)
}}}

This is looking good.  Hence I'd say this is fine to merge.

Is there anything I can do to get this into a release as fast as
possible such that ESO's notebooks work again?

andamian:
@msdemlei - thanks for the fix. We'll try to do the release in the next day or so.

PR #303, opened 2022-02-18
Add missing code and data files back to dist wheels.

tomdonaldson:
With the APE-17 changes (#290), wheel files built for release did not contain the code or test data.  @pllim and I found some subtle changes needed in `setup.cfg` to get those files in the wheel.

For the code, the key line is 
```
packages = find:
```
I also included:
```
zip_safe = False
```
which I don't totally understand, but seems OK after reading [the documentation](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html#setting-the-zip-safe-flag).

The data files had been handled by `setup_package.py` files.  With APE-17, these are obsolete so I deleted them and added their info to `setup.cfg` as `options.package_data` .

With these changes in place I verified that the wheel file created by these commands contained the same files as for the previous release (v1.2.1):
```bash
python -m pip install build wheel "twine>=3.3" -U
python -m build --sdist --wheel .
```

I also verified that the wheel was pip installable, and that the installation passed basic tests:
```bash
conda create --quiet --yes -n pyvo_install_test python=3.8
conda activate pyvo_install_test

pip install pyvo-1.2.<whatever>-py3-none-any.whl
pip install -U pytest-astropy requests_mock pillow

# Run local tests
cat << EOF | python
import pyvo
pyvo.test()
EOF

# Run one sample remote test (should have a few lines of output):
cat << EOF | python
from pyvo import registry
caoms = registry.search(keywords=['caom stsci'], servicetype='tap')
results = caoms[0].search(query='SELECT TOP 5 * FROM dbo.CaomObservation')
print(results)
EOF
```


pllim (approved):

pllim:
LGTM. Thanks!

tomdonaldson:
I expect the devastropy check will be fixed by astropy/astropy#12875.

pllim:
Re: devastropy -- yup yup, you can ignore it here.

Not sure if you are waiting for other reviews or not, but feel free to merge if you are not. Good work. Thanks!

bsipocz (approved):
LGTM

Issue #320, opened 2022-05-04
Datalinks in SCS not working correctly


zoghbi-a:
To reproduce the error:
```python
url = 'https://heasarc.gsfc.nasa.gov/xamin/vo/cone?showoffsets&table=numaster&'
res = pyvo.conesearch(url, (299, 35), radius=0.5)
res[0].getdatalink()
```
gives
```
AttributeError: 'SCSResults' object has no attribute '_adhocservices'
```

It looks like the order of the parent classes in `SCSResults` is culprit. As a test (using `res` from above):

```python
from pyvo.dal import DALResults
from pyvo.dal.adhoc import DatalinkResultsMixin

class SCSResults(DALResults, DatalinkResultsMixin):
    pass

class SCSResults_new(DatalinkResultsMixin, DALResults):
    pass

scs = SCSResults(res.votable)
scs_new = SCSResults_new(res.votable)

hasattr(scs '_adhocservices')  # -> False
hasattr(scs_new, '_adhocservices') # -> True

```

This makes `res[0].getdatalink()` above works. The other classes (`sia`, `sia2` and `ssa` seems to be ok).

andamian:
I'm seeing this potential problem (base class in the wrong spot) in other classes: `SIAService`, `SIAQuery` etc. Shouldn't those be fixed as well or it's considered too risky for the potential benefit?

msdemlei:
On Thu, May 05, 2022 at 03:58:23PM -0700, Adrian wrote:
> I'm seeing this potential problem (base class in the wrong spot) in
> other classes: `SIAService`, `SIAQuery` etc. Shouldn't those be
> fixed as well or it's considered too risky for the potential
> benefit?

For all I know and can see from the code, datalink ought to work just
fine on these.  Do you have a case where it's not?

PR #321, opened 2022-05-04
Fixes datalinks in scs (#320) by changing order of parent classes in dal.scs.SCSResults

zoghbi-a:
The order of inheritance is changes for SCSResult to fix issue #320.


msdemlei (approved):
This fixes a reasonably pertinent problem, and I can't see where it would cause problems otherwise.  I'd say let's go.

PR #326, opened 2022-05-09
Adding a local copy of the datalink vocabulary.

msdemlei:
This is used to make the datalink semantics tests local (which may
be a model for how other things using IVOA vocabularies might make
their tests non-remote).

This fixes bug #325


bsipocz (approved):
This indeed fixes the datalink test issues that surfaced up in #317

andamian (approved):

andamian:
Thanks you @msdemlei

PR #327, opened 2022-05-11
Adding session to SSAService and passing it through DatalinkService.

trjaffe:
A cleaner PR than #293 to allow session to be passed through  in SSA and DataLink as mentioned in issue #292

### DIFF
--- a/pyvo/dal/scs.py
+++ b/pyvo/dal/scs.py
@@ -413,7 +413,7 @@ class SCSQuery(DALQuery):
         return SCSResults(self.execute_votable(), url=self.queryurl, session=self._session)
 
 
-class SCSResults(DALResults, DatalinkResultsMixin):
+class SCSResults(DatalinkResultsMixin, DALResults):
     """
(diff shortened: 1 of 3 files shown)


andamian (approved):

pyvo/dal/scs.py
```diff
@@ -413,7 +413,7 @@ def execute(self):
         return SCSResults(self.execute_votable(), url=self.queryurl, session=self._session)
 
 
-class SCSResults(DALResults, DatalinkResultsMixin):
```
tomdonaldson, line 416: This new inheritance ordering makes sense as it's more consistent with the other *Results classes, but I'm idly curious whether it was a style change or you found you needed this to achieve the desired behavior.

tomdonaldson (commented):
Thanks @trjaffe. This looks good to me as well.

pyvo/dal/scs.py
tomdonaldson: Ah never mind.  I see this is already mentioned in #292.

Issue #322, opened 2022-05-05
Warning on QUERY_STATUS=OVERFLOW


andamian:
A `TAP` query returns a `QUERY_STATUS='OVERFLOW'` when the number of rows in the result exceeds `MAXREC` (set by the caller or the service), or the execution duration exceeds a threshold in the server or, in case of a async query, when the size of the results exceeds an output limit. While the result status can be easily determined with `result.query_status`, this is not always obvious for the caller especially when the result is returned via an intermediate package.

I think a warning would be more appropriate in this case. The user has the option of ignoring it or not. I've recently debugged a user problem in which the async query returned a weird number of rows and the culprit turned out to be `OVERFLOW`. This could have been easily avoided with such a warning.

Opinions?

I haven't looked at the `QUERY_STATUS='ERROR'` case.

msdemlei:
On Thu, May 05, 2022 at 04:24:31PM -0700, Adrian wrote:
> Opinions?

Definitely +1 on issuing a warning.

> I haven't looked at the `QUERY_STATUS='ERROR'` case.

IIRC that's translated into an exception at least on sync.  I'd have
to look up myself what happens when in async, but I'm pretty sure I'd
have insisted on not siliently passing over errors in that case
either.

PR #329, opened 2022-05-13
Warn on DAL overflow

andamian:
Fix for #322 .


msdemlei (approved):
Looks basically fine too me.  A few nitpicks  that I'd rather see fixed/addressed before merging this:

* Quite a few of the test VOTables were re-written and thus have meaningless changes.  I'm not saying that's a drama, but saying a few words how these changes came about would be great.
* There's now an overflow warning in pyvo/dal/tests/test_datalink.py::test_datalink_batch, which is because the obscore test result (pyvo/dal/tests/data/datalink/datalink-obscore.xml) is overflowing on creation.  I'd be fine with both manually removing the INFO element and with swallowing the warning.  It shouldn't go unhandled, though.

andamian:
> 



> Looks basically fine too me. A few nitpicks that I'd rather see fixed/addressed before merging this:
> 
> * Quite a few of the test VOTables were re-written and thus have meaningless changes.  I'm not saying that's a drama, but saying a few words how these changes came about would be great.

I've run the script that generates these files with a different version of `astropy`(`votable`). Where should I add this comment?

> * There's now an overflow warning in pyvo/dal/tests/test_datalink.py::test_datalink_batch, which is because the obscore test result (pyvo/dal/tests/data/datalink/datalink-obscore.xml) is overflowing on creation.  I'd be fine with both manually removing the INFO element and with swallowing the warning.  It shouldn't go unhandled, though.
Good point. I've missed that.

andamian:
@msdemlei - I've opted for removing the `OVERFLOW` status from the test data since that was a byproduct of the way they were generated (a `SELECT TOP` would not have had those in).

pyvo/dal/query.py
```diff
@@ -376,9 +381,13 @@ def _findstatus(self, votable):
 
     def _findstatusinfo(self, infos):
         # this can be overridden to specialize for a particular DAL protocol
+        status = None
+        # return the last status to catch potential overflow or error sent
```
tomdonaldson, line 385: This is interesting.  I had to go back and look at the DALI standard to see that this would work to the `OVERFLOW` case (i.e., that `OVERFLOW` would come last).

pyvo/dal/tests/data/query/overflowstatus.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Produced with astropy.io.votable version 5.0.1
+     http://www.astropy.org/ -->
+<VOTABLE version="1.4" xmlns="http://www.ivoa.net/xml/VOTable/v1.3" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.ivoa.net/xml/VOTable/v1.3 http://www.ivoa.net/xml/VOTable/VOTable-1.4.xsd">
+ <RESOURCE type="results">
+  <INFO ID="QUERY_STATUS" name="QUERY_STATUS" value="OK"/>
+  <INFO ID="QUERY_STATUS" name="QUERY_STATUS" value="OVERFLOW"/>
```
tomdonaldson, line 7: In my (brief) reading of the DALI spec, it looks like this `OVERFLOW` `INFO` element should come after the `TABLE`.  It looks like it would be handled either way though.

tomdonaldson (commented):
This looks good to me.  Just the one question about the placement of the `INFO` element in the test data.

PR #331, opened 2022-05-27
Added Jobs top element to fix the warning

andamian:
UWS is not aware of the `<jobs>` top element and warns:
```
pyvo/dal/tests/test_tap.py::TestTAPService::test_get_job_list
[119](https://github.com/astropy/pyvo/runs/6629666069?check_suite_focus=true#step:5:120)
  /home/runner/work/pyvo/pyvo/pyvo/utils/xml/elements.py:306: UnknownElementWarning: None:2:0: UnknownElementWarning: Unknown element jobs
  ```
  
  This is the structure of the `jobs` document that `PyVO` should be able to successfully parse:
  
  ```
  <?xml version="1.0" encoding="UTF-8"?>
<uws:jobs xmlns:uws="http://www.ivoa.net/xml/UWS/v1.0" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1">
  <uws:jobref id="zy070e3nebjaw5ta">
    <uws:phase>PENDING</uws:phase>
    <uws:ownerId>21</uws:ownerId>
  </uws:jobref>
</uws:jobs>
```

This fixes this to eliminate the warning.

### DIFF
--- a/pyvo/io/uws/endpoint.py
+++ b/pyvo/io/uws/endpoint.py
@@ -8,7 +8,7 @@ from astropy.utils.xml.writer import XMLWriter
 from astropy.io.votable.util import convert_to_writable_filelike
 
 from ...utils.xml.elements import xmlattribute, parse_for_object
-from .tree import JobSummary, JobList
+from .tree import JobSummary, Jobs
 
 __all__ = ["parse_job", "parse_job_list", "JobFile"]
 
@@ -45,8 +45,8 @@ def parse_job_list(
     --------
     pyvo.io.vosi.exceptions : The exceptions this function may raise.
     """
-    return parse_for_object(source, JobList, pedantic, filename,
-                            _debug_python_based_parser)
+    return parse_for_object(source, Jobs, pedantic, filename,
+                            _debug_python_based_parser).jobs
 
 
 def parse_job(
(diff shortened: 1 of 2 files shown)


andamian:
@msdemlei - I think this is pretty straightforward but given the intricacies of the XML code that I'm not fully grasping yet there might be things I've missed. Please let me know if that is the case. Thanks

msdemlei:
On Fri, May 27, 2022 at 11:31:37AM -0700, Adrian wrote:
> @msdemlei - I think this is pretty straightforward but given the
> intricacies of the XML code that I'm not fully grasping yet there
> might be things I've missed. Please let me know if that is the
> case. Thanks

Full disclosure: I'm not particularly familiar with that code, and
I've never delved into the magic that maps class names to XML element
names (I probably should have a good look at that one day).

Having said that, I'm pretty sure rather than *add* the Jobs element
you should just rename JobList to Jobs in the three places where
pyvo has it.

UWS metions "joblist" only as an endpoint name; the schema doesn't
have a "jobList" or anything like that either.  I'd suggest it was
just an oversight that the class was called JobList (which probably
didn't even yield warnings in those days).

andamian:
> 

That was my first attempt - to modify the `JobList` class. That class however just works with the list of jobs (`jobrer` elements) and I couldn't get it to parse properly. My "inspiration" for this solution was from the `JobSummary` class and its `Parameters` attribute, both of which being elements with lists of subelements.

andamian:
Closed in favour of #332

Issue #350, opened 2022-08-15
"date" parameter does not work with SSA search for certain services


EfremBraun:
I've found another error relating in pyVO which only appears with the https://heasarc.gsfc.nasa.gov/xamin/vo/ssa?table=hitomaster&" service. Based on https://github.com/astropy/pyvo/issues/347, I suspect this is an issue with the service rather than with pyVO, but I wanted to check.

```
import pyvo as vo
from astropy.coordinates import SkyCoord
from astropy.time import Time
import astropy.units as u
 
size = u.Quantity(0.5, unit="deg")
time = Time(('2015-01-01T00:00:00', '2018-01-01T00:00:00'), format='isot', scale='utc')
 
# Example from https://pyvo.readthedocs.io/en/latest/dal/index.html#simple-spectrum-access
pos = SkyCoord.from_name('Eta Carina')
ssa_service = vo.dal.SSAService("http://www.isdc.unige.ch/vo-services/lc")
ssa_results = ssa_service.search(pos=pos, diameter=size, time=time)
table = ssa_results.to_table()
print(table)
 
# Example using HEASARC's Hitomi search
pos = SkyCoord.from_name('perseus cluster')
ssa_service = vo.dal.SSAService('https://heasarc.gsfc.nasa.gov/xamin/vo/ssa?table=hitomaster&')
ssa_results = ssa_service.search(pos=pos, diameter=size, time=time)
table = ssa_results.to_table()
print(table)
```

The two searches are identical except for the choice of SSA service. The former works, but the latter fails with the following error:
```
Traceback (most recent call last):
  File "/Users/edbraun/Documents/xslide-git-repos/xslidepy/xslidepy/test.py", line 17, in <module>
    ssa_results = ssa_service.search(pos=pos, diameter=size, time=time)
  File "/Users/edbraun/opt/anaconda3/envs/xslide/lib/python3.9/site-packages/pyvo/dal/ssa.py", line 213, in search
    return self.create_query(
  File "/Users/edbraun/opt/anaconda3/envs/xslide/lib/python3.9/site-packages/pyvo/dal/ssa.py", line 570, in execute
    return SSAResults(self.execute_votable(), url=self.queryurl, session=self._session)
  File "/Users/edbraun/opt/anaconda3/envs/xslide/lib/python3.9/site-packages/pyvo/dal/adhoc.py", line 111, in __init__
    super().__init__(votable, url=url, session=session)
[...]

PR #351, opened 2022-08-16
No longer formatting microseconds into SSA time literals

msdemlei:
This started as an attempt to fix bug #350; it turns out it doesn't do that and the heasarc SSA has other problems.  But SSA 1.1, sect. 4.1.1.4, essentially implies we shouldn't be producing microseconds, so by the golden rule of interoperability just let's not.

### DIFF
--- a/pyvo/dal/ssa.py
+++ b/pyvo/dal/ssa.py
@@ -503,8 +503,15 @@ class SSAQuery(DALQuery):
             except TypeError:
                 raise valerr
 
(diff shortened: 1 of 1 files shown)


tomdonaldson (approved):
Thanks, @msdemlei. This seems like a reasonable work around and looks good to me.  The tests for SSA are a bit sparse, but that was a preexisting condition and seems out of scope for this minor change.

PR #363, opened 2022-09-21
Changing registry/index.rst example to use short names.

msdemlei:
For one, this is the practice we show to people.  And then, given this is now tested, we'd like to make the output stable, which it certainly is not if we use numeric indices -- the order of results is undefined in SQL.

There are still some test failures in this branch; I argue they are unrelated to this change.  One example, the get_tables one, also still fails, which is the consequence of a VizieR bug they'll soon fix.


bsipocz (approved):
Thanks, @msdemlei, this all looks good. I go ahead with a quick merge as I expect I'll need to rebase my PR once this one is merged.

Issue #376, opened 2022-10-13
`AsyncTAPJob` raises exception when getting `phase` during connection problems or when job is removed


JannisNe:
My code [`timewise`](https://github.com/JannisNe/timewise) uses `AsynTAPJob.phase` to check if the job is done. If a timeout occurs while calling `AsynTAPJob._update()`, an `AttributeError` will be raised while trying to construct a `DALServiceError` through `DALServiceError.from_except()` (see the last couple of lines in the stack trace below). This is partly mentioned in #365. 
A solution could be to add a [`backoff`](https://pypi.org/project/backoff/) decorator to `AsynTAPJob._update()`

```
Traceback (most recent call last):
  File "/lustre/fs23/group/icecube/necker/anaconda3/envs/air_flares/lib/python3.10/site-packages/urllib3/connectionpool.py", line 449, 
in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/lustre/fs23/group/icecube/necker/anaconda3/envs/air_flares/lib/python3.10/site-packages/urllib3/connectionpool.py", line 444, 
in _make_request
    httplib_response = conn.getresponse()
  File "/lustre/fs23/group/icecube/necker/anaconda3/envs/air_flares/lib/python3.10/http/client.py", line 1374, in getresponse
    response.begin()
  File "/lustre/fs23/group/icecube/necker/anaconda3/envs/air_flares/lib/python3.10/http/client.py", line 318, in begin
    version, status, reason = self._read_status()
  File "/lustre/fs23/group/icecube/necker/anaconda3/envs/air_flares/lib/python3.10/http/client.py", line 279, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/lustre/fs23/group/icecube/necker/anaconda3/envs/air_flares/lib/python3.10/socket.py", line 705, in readinto
    return self._sock.recv_into(b)
  File "/lustre/fs23/group/icecube/necker/anaconda3/envs/air_flares/lib/python3.10/ssl.py", line 1274, in recv_into
    return self.read(nbytes, buffer)
  File "/lustre/fs23/group/icecube/necker/anaconda3/envs/air_flares/lib/python3.10/ssl.py", line 1130, in read
    return self._sslobj.read(len, buffer)
TimeoutError: The read operation timed out

During handling of the above exception, another exception occurred:
                                                                                                                             
Traceback (most recent call last):
[...]

PR #392, opened 2023-01-03
Avoid `AttributeError` in `DALServiceError.from_except`

JannisNe:
If the response is `None` we get the `AttributeError` from [L.187](https://github.com/astropy/pyvo/blob/2ada6363f2d873b104c53dc0c9f6e270688114c5/pyvo/dal/exceptions.py#L187) in `pyvo.dal.exceptions`.
This is related to #376 .

### DIFF
--- a/pyvo/dal/exceptions.py
+++ b/pyvo/dal/exceptions.py
@@ -181,12 +181,15 @@ class DALServiceError(DALProtocolError):
             except AttributeError:
                 response = None
             code = 0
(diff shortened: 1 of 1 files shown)


msdemlei (approved):
Looks good to me.  It certainly would be nice to have a test exercising that, but since it clearly improves current misbehaviour, I'd merge it as-is.

andamian (approved):

PR #415, opened 2023-01-27
Make all-vo example work

msdemlei:
Querying for dataproduct_type was a bad idea as we get back all kinds
of strings (and sometimes NULLs, too), which vstack() refuses to
merge.

We're now using a query with floats, which should be a lot less fragile.
At least for the first few services, this now runs.  It will probably
fail when a service returns NULLs for s_ra and s_dec, though.


bsipocz (approved):
Thanks @msdemlei!

I'll go ahead and rebase this to remove the changes already been merged in from your other PR, and also to workaround the totally unrelated test failure.

PR #436, opened 2023-03-15
DOC: Improvements on doc for TAP service and other minor changes

chmmao:
Greetings, 

I would like to add some extensions of the TAP doc from a beginner user perspective, to make it more detailed and easier to start with. 

Change suggestions:
1. overall extension of the documentation on how to use TAP service, e.g. sync and async query
2. add links to parameter/class description for each service protocol
3. add some doc strings for several functions
4. minor fixes on spelling

### DIFF
--- a/docs/dal/index.rst
+++ b/docs/dal/index.rst
@@ -414,6 +414,7 @@ Finally, tables and their content can be removed:
 
     >>> tap_service.remove_table(name='test_schema.test_table')
 
+For further information about the service's parameters, see :py:class:`~pyvo.dal.TAPService`.
 
 .. _pyvo-sia:
 
@@ -432,7 +433,7 @@ Simple Image Access
     referred to as datacubes, cube or image cube datasets and may be considered examples 
     of hypercube or n-cube data. PyVO supports both versions of SIA.
 
-    -- `Simple IMage Access <https://www.ivoa.net/documents/SIA/>`_
+    -- `Simple Image Access <https://www.ivoa.net/documents/SIA/>`_
 
 Basic queries are done with the ``pos`` and ``size`` parameters described in
 :ref:`pyvo-astro-params`, with ``size`` being the rectangular region around
@@ -474,6 +475,8 @@ Available values:
 
 This service exposes the :ref:`verbosity <pyvo-verbosity>` parameter
 
+For further information about the service's parameters, see :py:class:`~pyvo.dal.SIAService`.
+
 .. _pyvo-ssa:
 
 Simple Spectrum Access
@@ -507,6 +510,7 @@ SSA queries can be further constrained by the ``band`` and ``time`` parameters.
     ...     time=time, band=Quantity((1e-13, 1e-12), unit="meter")
     ... )
 
+For further information about the service's parameters, see :py:class:`~pyvo.dal.SSAService`.
 
 .. _pyvo-scs:
 
@@ -533,7 +537,8 @@ within a circular region on the sky defined by the parameters ``pos``
     >>> scs_srv = vo.dal.SCSService('http://dc.zah.uni-heidelberg.de/arihip/q/cone/scs.xml')
     >>> scs_results = scs_srv.search(pos=pos, radius=size)
 
-This service exposes the :ref:`verbosity <pyvo-verbosity>` parameter
+This service exposes the :ref:`verbosity <pyvo-verbosity>` parameter.
+For further information about the service's parameters, see :py:class:`~pyvo.dal.SCSService`.
 
 .. _pyvo-slap:
 
@@ -552,6 +557,7 @@ Simple Line Access
 This service let you query for spectral lines in a certain ``wavelength``
 range. The unit of the values is meters, but any unit may be specified using
 `~astropy.units.Quantity`.
+For further information about the service's parameters, see :py:class:`~pyvo.dal.SLAService`.
 
 Jobs
 ====
@@ -569,7 +575,7 @@ to run several queries in parallel from one script.
 .. note::
     It is good practice to test the query with a maxrec constraint first.
 
-When you invoke ``submit job`` you will get a job object.
+When you invoke ``submit_job`` you will get a job object.
 
 .. doctest-remote-data::
 
--- a/docs/dal/index.rst
+++ b/docs/dal/index.rst
@@ -176,11 +176,12 @@ mean G-band magnitude between 19 - 20:
     2180349528028140800  310.5233961869657   50.3486391034819            19.0
 
 If you know a TAP service's access URL, you can directly pass it to
-:py:class:`~pyvo.dal.TAPService` to obtain a service object (see
-:ref:`pyvo.registry <pyvo-registry>` for when you do not know this access URL). 
-Here we use the `GAVO DC TAP <http://dc.g-vo.org/tap>`_ service to demonstrate. 
-To perform a query using ADQL, the ``search()`` method is used (note 
-that this method performs the synchronous query by default). 
+:py:class:`~pyvo.dal.TAPService` to obtain a service object. 
+Sometimes, such URLs are published in papers or passed around through
+other channels. Most commonly, you will discover them in the VO 
+registry (cf. :ref:`pyvo.registry<pyvo-registry>`).
+
+To perform a query using ADQL, the ``search()`` method is used. 
 TAPService instances have several methods to inspect the metadata
 of the service - in particular, what tables with what columns are
 available - discussed below.  Note that for exploratory query
--- a/docs/dal/index.rst
+++ b/docs/dal/index.rst
@@ -151,22 +151,28 @@ Getting started
 
 Let's start with a quick example using PyVO to perform a TAP/ADQL query.
 For instance we want to retrieve 5 objects from the GAIA DR3 database, 
-showing their id, position and mean G-band magnitude.
+showing their id, position and mean G-band magnitude between 19 - 20.
 
 .. doctest-remote-data::
     >>> import pyvo as vo
     >>> tap_service = vo.dal.TAPservice("http://dc.g-vo.org/tap")
-    >>> tap_service.search("SELECT TOP 5 source_id, ra, dec, phot_g_mean_mag FROM gaia.dr3lite")
+    >>> tap_service.search(
+            """ SELECT TOP 5 
+            source_id, ra, dec, phot_g_mean_mag 
+            FROM gaia.dr3lite
+            WHERE phot_g_mean_mag BETWEEN 19 AND 20
+            ORDER BY phot_g_mean_mag
+            """)
     <Table length=5>
         source_id              ra                dec         phot_g_mean_mag
                             deg                deg               mag      
         int64             float64            float64           float32    
     ------------------- ------------------ ------------------ ---------------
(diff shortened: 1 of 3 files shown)


docs/dal/index.rst
```diff
@@ -146,10 +146,168 @@ Table Access Protocol
...
+registry (cf. :ref:`pyvo.registry<pyvo-registry>`).
+
+To perform a query using ADQL, the ``search()`` method is used. 
+TAPService instances have several methods to inspect the metadata
+of the service - in particular, what tables with what columns are
+available - discussed below.  Note that for exploratory query
+construction, you can also use interactive TAP clients such as
+TOPCAT_, which include table browsers.
+
+.. _TOPCAT: http://www.star.bris.ac.uk/~mbt/topcat/
```
bsipocz, line 191: This is not a how-to for TAP, but a doc page for pyvo, please remove the cross reference.

docs/dal/index.rst
```diff
@@ -146,10 +146,168 @@ Table Access Protocol
 
     -- `Table Access Protocol <https://www.ivoa.net/documents/TAP/>`_
 
+Getting started
```
bsipocz, line 149: There is already a getting started heading in this very same document, I strongly suggest not to repeat it (it also messes up the heading references)

docs/dal/index.rst
```diff
@@ -146,10 +146,168 @@ Table Access Protocol
...
+of the service - in particular, what tables with what columns are
+available - discussed below.  Note that for exploratory query
+construction, you can also use interactive TAP clients such as
+TOPCAT_, which include table browsers.
+
+.. _TOPCAT: http://www.star.bris.ac.uk/~mbt/topcat/ 
+
+To get an idea of how to write queries in ADQL, have a look at
+`GAVO's ADQL course`_; it is basically a standardised subset of SQL
+with some extensions to make it work better for astronomy.
```
bsipocz, line 195: The definition of ADQL is already provided in the first paragraph, no need for the second half of the sentence. 
I'm also on the fence about the ADQL tutorial, but if @msdemlei wants to have the link here, then it should be fine.

docs/dal/index.rst
```diff
@@ -146,10 +146,168 @@ Table Access Protocol
...
+
+.. _GAVO's ADQL course: https://docs.g-vo.org/adql 
+
+Synchronous vs. asynchronous query
```
bsipocz, line 199: Not sure whether this belongs to this exact document. Again, the opening paragraph mentions them, and they are well-defined in the specs, so it's not exactly pyvo's task to explain what they mean on the server side.

docs/dal/index.rst
```diff
@@ -146,10 +146,168 @@ Table Access Protocol
...
+``submit_job()`` method. This submits an asynchronous query without 
+starting it, it creates a new object :py:class:`~pyvo.dal.AsyncTAPJob`.
+
+.. doctest-remote-data::
+
+    >>> job.url
+    (...)
+
+The job URL mentioned before is available in the ``url`` attribute;
+you can use that URL later to resume the job even from clients like TOPCAT.
```
bsipocz, line 240: again, why is there a TOPCAT reference?

docs/dal/index.rst
```diff
@@ -146,10 +146,168 @@ Table Access Protocol
...
+    >>> import pyvo as vo
     >>> tap_service = vo.dal.TAPService("http://dc.g-vo.org/tap")
-    >>> tap_results = tap_service.search("SELECT TOP 10 * FROM ivoa.obscore")
+    >>> tap_service.search(
+    >>>        """ SELECT TOP 5 
+    >>>        source_id, ra, dec, phot_g_mean_mag 
+    >>>        FROM gaia.dr3lite
+    >>>        WHERE phot_g_mean_mag BETWEEN 19 AND 20
+    >>>        ORDER BY phot_g_mean_mag
+    >>>        """)
```
bsipocz, line 166: Fix multi-line syntax, ensure that all doctest run and pass with locally running ``pytest docs/dal/ --remote-data``

```suggestion
    ...        """ SELECT TOP 5 
    ...        source_id, ra, dec, phot_g_mean_mag 
    ...        FROM gaia.dr3lite
    ...        WHERE phot_g_mean_mag BETWEEN 19 AND 20
    ...        ORDER BY phot_g_mean_mag
    ...        """)
```

chmmao:
> Thanks for the PR!
> 
> I had a few comments, 3 of which I would like to see changed (fix the code examples, remove duplicated heading name and the references to topcat. Python users may very well not know topcat and may prefer not to use GUIs, so self-contained documentation is much nicer than referencing totally unrelated tools and their features), the rest are more comments that can go either way, especially if the maintainers disagree with me.

Thank you for your detailed review. I just fixed the three issues that you mentioned. For the ADQL reference I will wait for @msdemlei 's comments.

docs/dal/index.rst
msdemlei: Well, I give you the ADQL course shows its age, but frankly I think it would be a good idea to link to *some* accessible ADQL resource.  The ADQL spec definitely is a terrible place to send anyone but an implementor to...

So, I'm happy to yield to any other gentle ADQL intro, but we really should avoid sending folks new to SQL-like stuff to IVOA documentation, and not saying anything about "what's this ADQL that I'm supposed to enter" isn't a good option either.

Talking about which: I think we're not saying anything about TAP examples anywhere yet?  They're also great to get people over the first few metres (try the examples attribute of a TAPService object).  Would this fit here, in addition?

docs/dal/index.rst
```diff
@@ -146,10 +146,156 @@ Table Access Protocol
...
+Consider the following example for using TAP and ADQL, retrieving 5
+objects from the GAIA DR3 database, showing their id, position and
+mean G-band magnitude between 19 - 20:
+
 .. doctest-remote-data::
 
+    >>> import pyvo as vo
     >>> tap_service = vo.dal.TAPService("http://dc.g-vo.org/tap")
-    >>> tap_results = tap_service.search("SELECT TOP 10 * FROM ivoa.obscore")
+    >>> ex_query = """
```
andamian, line 158: The "explore" step is missing here, IMO. Where do the user find the name of the table and columns?

Some services have a `description` that can be used here (similar to the SIA service above)

Others (like this one) have examples:

```
>>> for e in tap_service.examples:
>>> ...     print(e['QUERY'])
```

One can find the names of the tables:
```
[t for t in tap_service.tables.keys()]
```

And the columns in a table:
```
tap_service.search('select * from gaia.dr3lite', maxrec=1).table.columns
```

chmmao: Sorry for the late reply and many thanks for the review. I have added your suggestions accordingly into that location.

docs/dal/index.rst
```diff
@@ -174,6 +174,31 @@ mean G-band magnitude between 19 - 20:
...
 
+To explore more query examples, you can try either the ``description`` 
+attribute for some services. For other services like this one, try 
+the ``examples`` attribute.
+
+.. doctest-remote-data::
+
+    >>> for e in tap_service.examples:
+    ...     print(e['QUERY'])   # doctest: +SKIP
+    <query examples>
```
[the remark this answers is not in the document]
chmmao: Thanks for the review. 

Actually, all the ``doctest: +SKIP`` I put in the doc are because of their output EITHER are temporarily i.e. might be different later in time, OR the outputs are simply too long (that the output can not be shown fully), as far as I understand from ``doctest``, it checks the code output 1 to 1 with the strings that I wrote in the doc.
In this case, since the output is related to the names of the tables, it can indeed be updated later in time, and the list of the names is also too long. I am not sure if there is an ambiguous way for ``doctest`` to solve this issue.  I might go back to doctest documentation and check further details.

And for the variable issues, thanks for that hint, sure I will fix it later asap.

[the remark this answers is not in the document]
chmmao: Ah also here, the output for the query examples are also too long, there are many examples shown in the output, I could try too write them 1 to 1 into the doc, if that is okay.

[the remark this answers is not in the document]
bsipocz: for long outputs ellipses would work, or you can rewrite the example of looking for one special case in there.

For variable outputs, `doctest: +IGNORE_OUTPUT` is way more preferable than skipping it altogether, as it at least runs the line and spots any syntax, etc issues even if it doesn't compare the expected and actual outputs.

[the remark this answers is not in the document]
chmmao: Thanks for the hints. I have changed the `doctest` flags accordingly. Also I have to leave one `docktest: +SKIP` to line 319: `job.fetch_result()`. This is because the actual query is not faster than `doctest`'s processing speed, so it would turn out an error that the result does not exist yet.
Furthermore I have optimised some code outputs so that `docktest: +ELLIPSIS` would not be needed.
And  `dockets: + IGNORE_OUTPUTS` is indeed used for variable outputs.

docs/dal/index.rst
```diff
@@ -146,10 +146,196 @@ Table Access Protocol
...
+may encounter query timeouts.  Also, many data providers impose
+rather strict limits on the runtime alotted to sync queries. 
+
+In asynchronous (“async”) mode, on the other hand, the client just
+submits a query and receives a URL that let us inspect the
+execution status (and retrieve its result) later.  This means that
+no connection needs to be held, which makes this mode a lot more
+robust of long-running queries.  It also supports queuing queries,
+which allows service operators to be a lot more generous with
+resource limits.
```
andamian, line 249: Async queries are not bullet proof either. Service operators might impose limits on the size of the result set which can cause the query to fail or have the result truncated. Also, results might not be available before execution completes (i.e. streaming).

andamian (approved):

msdemlei:
On Wed, May 17, 2023 at 10:28:55AM -0700, Adrian wrote:
> Async queries are not bullet proof either. Service operators might
> impose limits on the size of the result set which can cause the
> query to fail or have the result truncated. Also, results might not
> be available before execution completes (i.e. streaming).

Do you argue text to this effect should be added to the brief
discussion of async?  Me... I'd say both caveats are perhaps not
overly central in this overview: That there will be resource limits
is, I think, more or less expectable.  And it would be the
availability of partial results up front that would require
mentioning in my view (but then I'm an implementor for whom this kind
of thing is an actual effort...).

On the other hand, I'm strongly against including these points
either, except that I think "bullet proof" doesn't quite fit what we
ought to say (perhaps: "note that resource limits even apply to async
queries" or so, although that sounds a bit trite to me).  Perhaps
let's merge and you add the extra material in another PR?

andamian:
The API presents both options because there's not clear winner but it makes it difficult for the user to choose the rights one without supporting information on the benefits and drawbacks. I remember it was one of the questions I wondered myself when I first came across this years ago.

bsipocz:
Given the approval, the fact that the PR is clearly an improvement over what we have atm,  and the sentiment in `Perhaps
let's merge and you add the extra material in another PR?`, I will go ahead and merge this. Any further improvements is welcome in a follow-up PR.

Thank you @chmmao!

bsipocz:
@astropy/pyvo-maintainers - note to ourselves: older PRs like this one should be rebased before merge to get a more recent CI status, so to ensure we don't add a breakage into `main`.

PR #439, opened 2023-04-19
Deprecating ivoid2service.

msdemlei:
The problem with it is that when a resource has multiple capabilities (e.g., SCS and TAP, as is now true for the big majority of VO resources), its result is ill-defined.

One *could* add a parameter to select the kind of service desired, but that's complicated in detail.  Hence, we advise people to use normal registry search with an ivoid constraint.

This is supposed to address Bug #425.

### DIFF
--- a/pyvo/registry/regtap.py
+++ b/pyvo/registry/regtap.py
@@ -22,6 +22,7 @@ import os
 import warnings
 
 from astropy import table
+from astropy.utils.decorators import deprecated
 from astropy.utils.exceptions import AstropyDeprecationWarning
 
 from . import rtcons
@@ -895,6 +896,9 @@ class RegistryResource(dalq.Record):
         return res
 
 
+@deprecated("1.5", "ivoid2service does not work in the presence of"
+    " multiple capabilities.  Use"
+    " registry.search(ivoid=...)[0].get_service('capname') instead.")
 def ivoid2service(ivoid, servicetype=None):
     """
     return service(s) for a given IVOID.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,8 @@
 
 - Deprecate VOSI ``AvailabilityMixin``, this mean the deprecation of the
   inherited ``availability``, ``available``, and ``up_since`` properties
-  of DAL service classes, too. [#413]
+
+- Deprecating ``ivoid2service`` because it is ill-defined. [#439]
 
 
 1.4.1 (2023-03-07)


bsipocz (approved):
This will need a changelog entry, but otherwise all good. 

The only thing I wonder that why this doesn't trigger a CI failure, didn't we have test coverage for ivoid2service?

pyvo/registry/regtap.py
```diff
@@ -895,6 +896,9 @@ def get_tables(self, table_limit=20):
         return res
 
 
+@deprecated("1.5", "ivoid2service does not work in the presence of"
+    " multiple capabilities.  Use"
+    " registry.search(ivoid=...)[0].get_service('capname') instead.")
```
bsipocz, line 901: Maybe chop up the `message` arg to `message` and `alternative`
```suggestion
@deprecated("1.5", message="ivoid2service does not work in the presence of multiple capabilities",
            alternative=" registry.search(ivoid=...)[0].get_service('capname')")
```

msdemlei:
On Wed, Apr 19, 2023 at 10:16:42AM -0700, Brigitta Sipőcz wrote:
> > ***@***.***("1.5", "ivoid2service does not work in the presence of"
> +    " multiple capabilities.  Use"
> +    " registry.search(ivoid=...)[0].get_service('capname') instead.")
>
> Maybe chop up the `message` arg to `message` and `alternative`

I considered that but then read the argument description "alternative
function or class name", which to me didn't sound free-text-ish
enough to put in something very definitely not just a name.

If you tell me "trust me, it's ok", I'll make the change.  Do you?

msdemlei:
On Wed, Apr 19, 2023 at 10:14:19AM -0700, Brigitta Sipőcz wrote:
> @bsipocz approved this pull request.
>
> This will need a changelog entry, but otherwise all good.
>
> The only thing I wonder that why this doesn't trigger a CI failure,
> didn't we have test coverage for ivoid2service?

No -- which is part of the reason why I'd like to scupper it: it's
not only a bad API, it would also require a lot of scaffolding (bad)
or a remote test (worse) to test.

bsipocz:
> If you tell me "trust me, it's ok", I'll make the change.  Do you?

Actually, it's not OK. `alternative` accepts a free text, but unfortunately when `message` is provided, it overrides it all. The behaviour is somewhat different than what I would have expected. So what you have here is in fact better. 
I'm sorry for the noise.

In the meantime, merging the other PR introduced a changelog conflict, I'll go ahead and rebase and then merge this.

bsipocz:
Thank you Markus!

PR #482, opened 2023-08-23
Removing a remote-data mark in a RegTAP test.

msdemlei:
That was spurious in this position anyway; the fixture in question does not require network access (but the tests it is there for do, so they have the mark).

This fixes bug #480.


bsipocz (approved):
Thanks! Devdeps failure is upstream so I go ahead with merge

PR #506, opened 2023-12-15
Make registry.Interface construction work with no standard_id.

msdemlei:
Interface.__init__ so far tries a .startswith on something that can be None (actually, will be by virtue of the preceding line). This commit fixes this (at the cost of is_vosi being variously False or None, which should be operationally insignificant for now).

This is a really tiny fix -- perhaps we can sneak it in without a changelog entry?

### DIFF
--- a/pyvo/registry/regtap.py
+++ b/pyvo/registry/regtap.py
@@ -348,10 +348,14 @@ class Interface:
     def __init__(self, access_url, standard_id, intf_type, intf_role):
         self.access_url = access_url
         self.standard_id = standard_id or None
-        self.is_vosi = standard_id.startswith("ivo://ivoa.net/std/vosi")
         self.type = intf_type or None
         self.role = intf_role or None
         self.is_standard = self.role == "std"
+
+        if self.standard_id is not None:
+            self.is_vosi = self.standard_id.startswith("ivo://ivoa.net/std/vosi")
+        else:
+            self.is_vosi = False
 
     def __repr__(self):
         return (f"Interface({self.access_url!r}, {self.standard_id!r},"


pyvo/registry/regtap.py
```diff
@@ -348,7 +348,8 @@ class Interface:
     def __init__(self, access_url, standard_id, intf_type, intf_role):
         self.access_url = access_url
         self.standard_id = standard_id or None
-        self.is_vosi = standard_id.startswith("ivo://ivoa.net/std/vosi")
+        self.is_vosi = (standard_id
+            and standard_id.startswith("ivo://ivoa.net/std/vosi"))
```
tomdonaldson, line 352: Although I agree that there is unlikely to be a problem with `self.is_vosi` sometimes being `None` versus `False`, I do think there is some value in consistency, and alternative code not too verbose still easy to read (as long as we don't use the "ternary operator").
```suggestion
        if standard_id is not None:
            self.is_vosi = standard_id.startswith("ivo://ivoa.net/std/vosi")
        else:
            self.is_vosi = False
```

As a secondary comment, I was considering whether to ask for covering this with a unit test and looked at the existing tests.  It was curious to me why [this test](https://github.com/astropy/pyvo/blob/3c8c67fca08c9458bffe229674e39502489a6202/pyvo/registry/tests/test_regtap.py#L209) hasn't been raising exceptions when I realized that the modified statement above is using `standard_id` the function arg as opposed to `self.standard_id` the class member, so in the test `standard_id` is not `None` even though `self.standard_id` is.  So maybe a better suggestion would be this since it would then get tested for free with the existing test:
```suggestion
        if self.standard_id is not None:
            self.is_vosi = self.standard_id.startswith("ivo://ivoa.net/std/vosi")
        else:
            self.is_vosi = False
```

tomdonaldson:
> This is a really tiny fix -- perhaps we can sneak it in without a changelog entry?

I agree that no change log is needed for this.

msdemlei:
On Fri, Dec 15, 2023 at 07:15:11AM -0800, Tom wrote:
> @tomdonaldson commented on this pull request.
> there is some value in consistency, and alternative code not too
> verbose still easy to read (as long as we don't use the "ternary

Fair enough.  I've force-pushed something that takes your suggestion.

Do you let me get away without a changlog entry?  To me, it seems a
bit too trivial to mention, and it seems it's not bitten anyone yet.

tomdonaldson (approved):
Thanks @msdemlei.  This looks good to me.

PR #517, opened 2024-01-30
The regtap Ivoid constraint can now take multiple ivoids.

msdemlei:
As usual in the pyvo Regtap API, these will then be matched in a disjunction.


bsipocz (approved):
Thank you @msdemlei! I'll push a commit that fixes pep8 and will merge this.

PR #239, opened 2020-09-18
WIP: Move astropy.samp to PyVO

pllim:
Fix #155 .

This is the first step towards astropy/astropy#9763 . cc @astrofrog 

**TODO**

- [ ] Get buy-in and review from PyVO.
  - ~Need to wait for PyVO to drop Python 3.5 first.~ #255 
  - If PyVO applies APE 17, can undo the changes to `setup_package.py`.
- [ ] Port over typo fixes from astropy/astropy#11967
- [ ] Incorporate review comments.
- [x] Make sure tests pass.
- [x] Make sure render doc looks fine.
- [ ] Add change log? (Can't find a section for unreleased version.)

**Notes**

* I cannot name this new subpackage `pyvo.samp` because you already have `samp.py`. I don't know how it overlaps with `astropy.samp`.
* `astropy.samp` used `astropy.__version__`, this needs to change now. Needs discussion.
* This subpackage uses `astropy.config` system. I don't think we can change it for the sake of backward compatibility.
* This subpackage uses `astropy.log` for logging. If PyVO has its own logging, it can be replaced.
* There is usage of hidden function `astropy.config.path._find_home`. If you have a better replacement, please let me know.
* `astropy` moved away from a lot of the `astropy_helpers` setup stuff. I tried to move them back here (e.g., `setup_package.py`) since you have not implemented APE 17 yet. Need extra pairs of eyes to make sure I didn't mess things up.
* Replaced `astropy.samp` with new import.

**Disclaimer:** I don't maintain `astropy.samp`, Tom Robitaille does.

PR #527, opened 2024-02-20
Move astropy.samp to pyvo.astropy_samp

### LINKED ISSUE #155: Move astropy.samp to pyvo?
Did we discuss this at IVOA Interop? cc @astrofrog

MridulS:
To fix #155 and another take on https://github.com/astropy/pyvo/pull/239


bsipocz:
@astropy/coordinators -- how is Mridul not yet part of the org? Please resolve it, so we don't need to approve CI on his PRs.

pllim:
`E           astropy.samp.errors.SAMPProxyError: <SAMPProxyError 2: 'Client cli#2 not subscribed to MType table.load.votable'>`

Not sure what is going on but @tomdonaldson and/or @astrofrog would be the best person to advice for this effort.

`astropy` already dropped Python 3.8 but hopefully should not matter here since we rarely touch `astropy.samp`.

ManonMarchand:
It looks like the naming convention for the modules here is `pyvo.group.standard` or `pyvo.standard` which would give `pyvo.app.samp` or `pyvo.samp`. Which one do you all think is best ?

Also I cannot run the tests from this branch. It looks like they are disabled since this https://github.com/astropy/astropy/commit/7c7e48836c84685ee53c0e785547960db959a490 ?

msdemlei:
On Mon, Apr 15, 2024 at 02:07:09AM -0700, Manon Marchand wrote:
> It looks like the naming convention for the modules here is
> `pyvo.group.standard` or `pyvo.standard` which would give
> `pyvo.app.samp` or `pyvo.samp`. Which one do you all think is best?

I'm all for short paths, so it would be pyvo.samp for me.

There is a bit of trouble there because pyvo.samp already contains a
few convenience wrappers, but from the top of my head I don't think
there are name clashes between the old astropy samp namespace and the
old pyvo.samp.

pllim:
> I cannot run the tests from this branch

Feel free to disable them in your branch here.

p.s. Though by now I think Mridul's contract is done and he has moved on, so maybe a `pyvo` maintainer should take this over.

pllim:
It is not up to me but I would be happy if someone can take this over and push it over the finish line. Likely you have to start fresh because there are new changes to `astropy.samp` upstream since, but the diff here hopefully can still inspire you.

PR #565, opened 2024-06-27
Fixing the TAP doctest in dal/index.rst.

msdemlei:
The TOP 10 * that used to be in the query is highly dependent on the state of the remote machine, so it's a miracle it worked as well as it did.

The query I am using now should be stable for the forseeable future (CALIFA is done).

I have, for now, dropped the example for getdataobj().  The output of that is extremely opaque, and I'm not sure I'm a big fan of that method in the first place.  *If* we show it, we have to say something about what the intended use is -- in this case, this could be a brief excursion into datalink (retrieving a #preview, perhaps?), but I'd say that's for another day.

closes #564


bsipocz (approved):
thank you!

Issue #575, opened 2024-07-11
Table utypes are not exposed in RegistryResource.get_tables()


msdemlei:
If you try:

```
rscs = pyvo.registry.search(
  ivoid="ivo://org.gavo.dc/__system__/obscore/obscore")
print(rscs[0].get_tables()["ivoa.obscore"].utype
```

you will get an AttributeError; this is against the same functionality from VOSI, where the utype is preserved, and it is, in addition, inconvenient because the utype is relevant to discover tables conforming to obscore (well: in the future), EPN-TAP, LineTAP, and others.

Let's have utype in the registy table stubs, too.

PR #576, opened 2024-07-11
RegistryResults.get_tables now preserves table utypes.

msdemlei:
This is supposed to fix bug #575.

In addition, while creating a test for this, I noticed that pytest has scope="function" for fixtures by default.  I take the liberty of sneaking in scope="session" for all the regtap fixtures, which speeds up the tests between a bit and (for remote fixtures) dramatically.


ManonMarchand:
Looks good to me :slightly_smiling_face:

ManonMarchand (approved):

PR #580, opened 2024-07-19
SodaRecordMixin no longer will use a datalink#links endpoint for soda

### LINKED ISSUE #578: Confusing behaviour with SODA .processed when no in-response block is present
Consider the following code that does a spectral cutout from a cube found in a TAP service:

```
import pyvo
from astropy import units as u

svc = pyvo.dal.TAPService("http://dc.g-vo.org/tap")
res = svc.run_sync(
  """SELECT accref, mime FROM califadr3.cubes
  WHERE target_name='IC1151' AND setup='COMB'""")
for row in res:
  with open("res.fits", "wb") as f:
    f.write(row.processed(band=(600*u.nm, 620*u.nm)).read())
```

This works (try it: it's a fairly pretty galaxy, and it's fast, too).

However, just replacing the ADQL select list, like this:

```
import pyvo
from astropy import units as u

svc = pyvo.dal.TAPService("http://dc.g-vo.org/tap")
res = svc.run_sync(
  """SELECT * FROM califadr3.cubes
  WHERE target_name='IC1151' AND setup='COMB'""")
for row in res:
  with open("res.fits", "wb") as f:
    f.write(row.processed(band=(600*u.nm, 620*u.nm)).read())
```
the following extremely confusing error message results:

```
pyvo.dal.exceptions.DALServiceError: UsageError: Field band: The following parameter(s) are not accepted by this service: band
 for http://dc.zah.uni-heidelberg.de/califa/q3/dl/dlmeta
```

Let me try to untangle this.

The failure in the second case results from this code in dal/adhoc.py:

```
[_get_soda_resource]
        # let it count as soda resource
        try:
            return self._results.get_adhocservice_by_ivoid(
                "ivo://ivoa.net/std/datalink#links")
        except DALServiceError:
            pass
```

This falls back to a links resource when there is no SODA resource.  I think this is simply wrong; I can't see any reason why we ought to expect that such a links resource (reminder: this is supposed to return a datalink document) should understand any of SODA's parameters, let alone do something sensible with them.  Sure enough, this datalink links service doesn't know about BAND, and that's where the message comes from.

This code was contributed by @funbaker in commit afa2b919 with the comment "combine and optimize datalink and soda".  So, my hunch is that this code's intention was something that it simply does not do.  Or does someone see a case in which this could be useful after all?

I suggest to remove it; without it, the first code succeeds, too.

And then there is the interesting question why the second code succeeds even with current pyVO.  Well, the success is because with accref and mime our datalink code has enough to figure out that the *product* described in the row comes in datalink and duly retrieves it.

But accref and mime surely is in * as well, you say.  True.  But the generating service knows that these two columns are not good enough to operate the datalink service associated with this table, as its ID is filled from the obs_publisher_did column.   The service hence does not include a block for the links service, and pyVO does not trip before it notices that the access URL already points to the datalink document it would get from a links endpoint.

Selecting obs_publisher_did on top of accref and mime will trigger the problem.

Yes: it took me a while to figure this puzzle out.

Summing up:  Let's drop the "let it count as soda resource" thing.  It's an accident waiting to happen sometimes immodest enough to actually happen in confusing ways.

msdemlei:
It is unclear why it ever did this, as I really do not see how this would ever have a beneficial effect (the only parameter understood by both links and SODA is ID, and even that is interpreted in totally different ways).  It has, however, a detrimental effect; see bug #578.

Fixes #578

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,6 +49,8 @@ Enhancements and Fixes
 Deprecations and Removals
 -------------------------
 
+- SodaRecordMixin no longer will use a datalink#links endpoint for soda [#580]
+
 
 1.5.2 (2024-05-22)
 ==================
--- a/pyvo/dal/adhoc.py
+++ b/pyvo/dal/adhoc.py
@@ -671,13 +671,6 @@ class SodaRecordMixin:
         except DALServiceError:
             pass
 
-        # let it count as soda resource
-        try:
-            return self._results.get_adhocservice_by_ivoid(
-                "ivo://ivoa.net/std/datalink#links")
-        except DALServiceError:
-            pass
-
         dataformat = self.getdataformat()
         if dataformat is None:
             raise DALServiceError(


andamian:
From what I've searched, I can only imagine that the intend might have been to try other `datalink` links endpoints with different authorizations. Even so, those datalink links were not resolved so I don't know how the code might have worked. I'm fine with removing it.

andamian (approved):

PR #622, opened 2024-11-21
Add dot to extensions of added MIME types

hugovk:
There's a suggestion in https://github.com/python/cpython/issues/75223 / https://github.com/python/cpython/pull/2895 to raise a `ValueError` when adding MIME type extensions with a dot.

This PR adds the dots.


bsipocz (approved):
Thank you for the PR, it looks uncontroversial, so I go ahead with a merge.

PR #659, opened 2025-03-20
More careful NULL value handling in tapregext data limits.

msdemlei:
This ought to fix bug #658.

Actually, I considered including the (perhaps more relevant) default upload limit in the describe() output, too.  I will if someone asks for it here...

### DIFF
--- a/pyvo/io/vosi/tapregext.py
+++ b/pyvo/io/vosi/tapregext.py
@@ -369,6 +369,9 @@ class DataLimit(ContentMixin, Element):
 
         self.unit = unit
 
+    def __str__(self):
+        return f"{self.unit}:{self.content}"
+
     @xmlattribute
     def unit(self):
         return self._unit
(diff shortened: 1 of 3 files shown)


ManonMarchand:
I think default and hard limits are both interesting in the description.

ManonMarchand:
This explains why the two values are identical in the capabilities endpoints I know! I guess the services can add warnings in the response when people overshoot the default limit so we don't need it in pyVO. And it it very strange from IRSA not to set an upload limit at all.

The PR fixes a very rare issue. It can go without a changelog entry. Could you maybe add a test in https://github.com/astropy/pyvo/blob/main/pyvo/io/vosi/tests/test_capabilities.py ?

msdemlei:
On Thu, Mar 20, 2025 at 06:27:17AM -0700, Manon Marchand wrote:
> The PR fixes a very rare issue. It can go without a changelog
> entry. Could you maybe add a test in
> https://github.com/astropy/pyvo/blob/main/pyvo/io/vosi/tests/test_capabilities.py
> ?

Fair enough, in particular because test coverage here is fairly
spotty (there's no coverage of tapregext's describe, to begin with,
but then the way these describe-s are done isn't very test-friendly).

I've added a few tests that should fail when we assume too much about
the content of TAPRegExt.  We certainly could do better still, but I
think I have covered a few spots that might blow up in the future.

bsipocz (approved):
Thank you so much for having this fixed so quickly!

Issue #676, opened 2025-05-31
UWS jobInfo parser generates warnings for UWS jobs that include a jobInfo element


stvoutsin:
Consider this example:

        <uws:job xmlns:uws="http://www.ivoa.net/xml/UWS/v1.0" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1">
            [Rest of job omitted]
            <uws:jobInfo>
                <tapQueryInfo>
                    <pct_complete>100</pct_complete>
                    <chunks_processed>1</chunks_processed>
                    <total_chunks>1</total_chunks>
                </tapQueryInfo>
            </uws:jobInfo>
        </uws:job>

If this is encountered as the output of a UWS job, pyvo will produce the following warnings:

      WARNING: UnknownElementWarning: None:22:8: UnknownElementWarning: Unknown element jobInfo [pyvo.utils.xml.elements]
      WARNING:astropy:UnknownElementWarning: None:22:8: UnknownElementWarning: Unknown element jobInfo
      WARNING: UnknownElementWarning: None:23:8: UnknownElementWarning: Unknown element tapQueryInfo [pyvo.utils.xml.elements]
      WARNING:astropy:UnknownElementWarning: None:23:8: UnknownElementWarning: Unknown element tapQueryInfo
      WARNING: UnknownElementWarning: None:24:8: UnknownElementWarning: Unknown element pct_complete [pyvo.utils.xml.elements]
      WARNING:astropy:UnknownElementWarning: None:24:8: UnknownElementWarning: Unknown element pct_complete
      WARNING: UnknownElementWarning: None:25:8: UnknownElementWarning: Unknown element chunks_processed [pyvo.utils.xml.elements]
      WARNING:astropy:UnknownElementWarning: None:25:8: UnknownElementWarning: Unknown element chunks_processed
      WARNING: UnknownElementWarning: None:26:8: UnknownElementWarning: Unknown element total_chunks [pyvo.utils.xml.elements]
      WARNING:astropy:UnknownElementWarning: None:26:8: UnknownElementWarning: Unknown element total_chunks
      WARNING: UnknownElementWarning: None:28:30: UnknownElementWarning: Unknown element jobInfo [pyvo.utils.xml.elements]

The context of this is that we're prototyping how to add progress information for a UWS job, and the jobInfo element seemed like a good option for this.

The UWS spec here states:

[...]

PR #679, opened 2025-06-04
fix: change JobSummary (UWS) to support jobInfo elements with any XML content

stvoutsin:
### Problem

Pyvo currently doesn't support `jobInfo` elements without triggering warnings.
https://github.com/astropy/pyvo/issues/676

### Solution in PR

Add support for arbitrary XML content in UWS jobInfo elements without generating warnings.
Implements ExtensibleUWSElement that stores unknown child elements in a dictionary, enabling access to any custom jobInfo content.

### Tests

Added tests to verify behavior with various scenarios of valid UWS jobs with jobInfo elements.

### Potential Issues

- If there are tag-name collisions in the `jobInfo` element with different (or no) namespaces, this will end up using the last one. A different approach if we are worried about this use case would be to store the namespace information along with the content. The one issue there is that `parse` strips the namespace from the element (Although perhaps it can be found in `data`?). If we do want to support this use case I can make amendments to the PR

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@
 Enhancements and Fixes
 ----------------------
 
+- Add support for the jobInfo element for UWS jobs [#679]
 
 Deprecations and Removals
 -------------------------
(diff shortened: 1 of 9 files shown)


msdemlei (approved):
I think I by and large the API, although I'm not sure I like the magic in .value vs. .text... it doesn't feel *quite* right, and I'm sure people will write stuff like job.jobinfo.get("status").value, which would yield an AttributeError where a KeyError would be much more appropriate.  Hm.  But then we should probably tell people to write self.jobinfo["status"] anyway, and then the problem goes away.

I'd say let's merge with the proviso that it might still change if people don't like it.

Oh, and another thing: Please promise you'll write documentation on this.  docs/io/uws.rst is mainly a stub right now, and it would be a dream if this were to be come a reason to fill in a bit more about what UWS is, when you'll see it, and what jobInfo is.

If you start it, I promise I'll try things on my indexing UWS jobs and write something about that, too.

Issue #715, opened 2026-01-12
session not passed through to datalink query object


trjaffe:
Like issue #446 , when debugging a service that is on a development server that doesn't have a valid certificate, we need to be able to specify that we don't care.  Otherwise we cannot debug.  This is still the case with DatalinkService.  The trick tried below works for a development TAP service but not the DataLink service.  Code below works only if you change line 512 of _pyvo/dal/adhoc.py_ to include `, session=self._session`.  Unfortunately, you'll have to find your own server that hasn't got a certificate if you want to test, since ours is behind a firewall.  

```
import pyvo as vo
import requests
session=requests.Session()
session.verify=False
import warnings
# we know
warnings.filterwarnings("ignore", module="urllib3.*")

ids=["ivo%3A%2F%2Fnasa.heasarc%2Fswiftmastr%3F00036555179%2Fswift.obs.xrt.prod.skimg&","ivo%3A%2F%2Fnasa.heasarc%2Fswiftmastr%3F00036555179%2Fswift.obs.xrt.prod.imgql&"]
try:
    dlservice = vo.dal.adhoc.DatalinkService(\
        "https://my.service/vo/datalink?",session=session)
    query = dlservice.create_query(id=ids)
    links = query.execute(post=True)
    print("********* devserver run with unverified session works")
except Exception as e:
    print(f"********* devserver run with unverified session gives exception:  {e}")
```

Trivial PR coming.

PR #716, opened 2026-01-12
Add session parameter to DatalinkQuery initialization

trjaffe:
For issue https://github.com/astropy/pyvo/issues/715, this tiny change ensures the session parameters get down to the request used.


bsipocz:
I would call it a bug fix and will include it in the next fix release if we have one as we should have passed the session along already.

PR #718, opened 2026-01-23
Add DALRateLimitError for HTTP 429 responses with retry timing info

### LINKED ISSUE #717: Improving handling of 429 (Rate limit) responses in DAL
## Summary

When a DAL service returns HTTP 429 (i.e. Rate limits), PyVO raises a
generic `DALServiceError` with no structured way to access retry timing information.
If a users wants to determine when they can retry, they would currently have
to parse the response object.

Example of output from a rate-limited request currently:
`pyvo.dal.exceptions.DALServiceError: 429 Client Error: None for url: https://example.com/tap/async/phase
`
### Proposed Solution

Add `DALRateLimitError` subclass of `DALServiceError` with retry after fields

### Desired API

```
try:
    results = tap.run_async(query)
except DALRateLimitError as e:
    print(f"Rate limited. Retry after {e.retry_after_seconds}s")
```

I've drafted what this might look like here:
https://github.com/astropy/pyvo/compare/main...stvoutsin:pyvo:handle-rate-limits
If this approach makes sense I'm happy to put in a PR

stvoutsin:
## Summary
When a DAL service returns HTTP 429 (i.e. Rate limits), PyVO raises a
generic DALServiceError with no structured way to access retry timing information.
If a users wants to determine when they can retry, they would currently have
to parse the response object.

## Fix
Add DALRateLimitError subclass of DALServiceError with retry after fields

## Usage
```
try:
    results = tap.run_async(query)
except DALRateLimitError as e:
    print(f"Rate limited. Retry after {e.retry_after_seconds}s")
```

This gives the users a better indication on the rate limits and gives them the option to better automate the process of querying and retrying based on the structured information.

## Relevant Issue
This addresses issue https://github.com/astropy/pyvo/issues/717

## Other considerations
I also considered whether it is worth adding an automated retry mechanism (either by default or optional), but figured that this may lead to unexpected and non-obvious behavior and perhaps we should leave it up to the user to retry and give them better clarity on what is happening.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ Enhancements and Fixes
 
 - Pass session through to DatalinkService requests (#715)
 
+- Add DALRateLimitError for HTTP 429 responses with retry timing info [#718]
+
 
 Deprecations and Removals
 -------------------------
--- a/pyvo/__init__.py
+++ b/pyvo/__init__.py
@@ -39,4 +39,4 @@ from .registry import search as regsearch
 from .dal import (
     imagesearch, spectrumsearch, conesearch, linesearch, tablesearch,
     DALAccessError, DALProtocolError, DALFormatError, DALServiceError,
-    DALQueryError)
+    DALQueryError, DALRateLimitError)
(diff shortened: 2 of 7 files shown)


bsipocz (approved):
Generally looks good to me, some nitpicking comments otherwise I think we merge it. Thank you.

Issue #713, opened 2025-12-11
Intermittent ReadTimeout failures in pyvo async TAP polling


jkrick:
### Description

When using PyVO’s asynchronous TAP workflow (submit_job() + polling of job.phase), I see intermittent ReadTimeout errors during the job-phase polling step. Synchronous (run_sync) queries do not exhibit this behavior at all.

Environment:
PyVO version: 1.8
Python: 3.12.3

```
service = vo.dal.TAPService("https://irsa.ipac.caltech.edu/TAP")
```

Query: modest cone search on CosmoDC2 (10–30 seconds typical runtime)

The most revealing test I did:
I benchmarked the exact same ADQL query 20 times in two modes:
1. Using run_sync()
20/20 successes
Runtime variable (2–30 seconds) but no timeouts, no failures

2. Using async workflow with service.submit_job(adql), job.run() and job.phase
Only ~70–95% success rate
Failures all have the same signature (below).

I don't know what is happening, but it is certainly difficult to make repeatable.  ChatGPT suggests that it is the hard-coded 10s read timeout being triggered during a job status request, and not something about the TAP server or the underlying database.  Evidence for this is the successful .run_sync queries. 

minimum failing example and full traceback below.  

```
import time
import pandas as pd
import pyvo

# Setup TAP service (CosmoDC2 mock catalog at IRSA)
service = pyvo.dal.TAPService("https://irsa.ipac.caltech.edu/TAP")
tablename = 'cosmodc2mockv1_heavy'

# ADQL query to benchmark
adql = f"""
SELECT redshift
FROM {tablename}
WHERE CONTAINS(
    POINT('ICRS', ra, dec),
    CIRCLE('ICRS', 54.0, -37.0, 0.05)
) = 1
"""

# ----------------------------------------------------
# Benchmark async job behavior
# ----------------------------------------------------
N = 30
records = []

for i in range(N):
    start = time.time()

    try:
        job = service.submit_job(adql)
        job.run()

        # Poll .phase until COMPLETED, ERROR, or timeout
        timeout = 60
        poll_interval = 5
        t0 = time.time()

        while True:
            phase = job.phase  # <-- sometimes triggers ReadTimeoutError internally

            if phase == "COMPLETED":
                result = job.fetch_result()
                runtime = time.time() - start
                records.append(dict(trial=i+1, runtime_s=runtime,
                                    success=True, error=None))
                break

            if phase in ("ERROR", "ABORTED", "UNKNOWN"):
[...]

PR #721, opened 2026-01-29
Add DEFAULT_JOB_POLL_TIMEOUT constant and use it in _update method

stvoutsin:
## Description

This is an approach to somewhat address issue https://github.com/astropy/pyvo/issues/713

The proposed changes here add a `DEFAULT_JOB_POLL_TIMEOUT` module-level constant and updates the `_update()` method to  use that when the timeout parameter is not provided.

I have not modified the default value to > 10 seconds because I think we concluded that that anything more than that in these requests indicates a service-level problem.

### Usage

For users who want to overwrite this, they can use it as:
```
import pyvo.dal.tap
pyvo.dal.tap.DEFAULT_JOB_POLL_TIMEOUT = 30
service = pyvo.dal.TAPService("https://irsa.ipac.caltech.edu/TAP")
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 Enhancements and Fixes
 ----------------------
 
+- Add DEFAULT_JOB_POLL_TIMEOUT constant [#721]
+
 - Pass session through to DatalinkService requests (#715)
 
 - Add DALRateLimitError for HTTP 429 responses with retry timing info [#718]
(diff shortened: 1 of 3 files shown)


bsipocz (approved):
Thank you!

I wonder if something on the instance level might be better so we could set the timeout to be different depending on the services -- but also, this is good enough as is already. 

Let's see what the other maintainers think though.

bsipocz:
(I'll deal with the unrelated leapsecond issue separately)

PR #723, opened 2026-02-02
Install VOTable schemas

McSinyx:
Follow up on GH-698, bug was [found while packaging version 1.8](https://codeberg.org/guix/guix/issues/4393).

Fixes: fd700322edad ("Use of local copies of VOTable schemas")


bsipocz (approved):
Good catch, thank you for the fix!

bsipocz:
I'm planning to tag a 1.8.1 release, possibly later this week, in case you prefer to wait rather than to patch for packaging.

PR #726, opened 2026-02-04
Update the observation date modeling that has been changed during the

lmichel:
During the MANGO REC process, it has been suggested to simplify the representation of the observation dates (`mango:EpochPostion.obsDate`).
The `mango:ObsDate` complex type has been replaced with a simple attribute where the representation is given by its dmtype (`<ATTRIBUTE dmrole="mango:EpochPosition.obsDate" dmtype="year" value="2000.0" \>`).

This change requires the `SkyCoord` builder to be updated as well as the annotation builder.


lmichel:
After the last comments in the MANGO data model RFC suggesting changes in the the observation date representation, this PR must be delayed until the new code support the final time representation in the model.

lmichel:
I found out a mismatch in the `Julian/Besselian` year types supported by `mangoObsDate` .
- `JulianYear` instead of `JulianEpoch` in the model
- `BesselianYear` instead of `BesselianEpoch` in the model

I've updated the code accordingly

tomdonaldson (approved):
Thanks @lmichel.  This looks good to me.  The test cases were very helpful.

PR #727, opened 2026-02-04
Fix sia2 to not display warnings when user-provided maxrec use case

stvoutsin:
## Description

Modify SIA2 execute to check for overflows via `check_overflow_warning` before returning the result

This closes issue #728

## Tests
Added unit tests for testing various scenarios of use with maxrec

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 Enhancements and Fixes
 ----------------------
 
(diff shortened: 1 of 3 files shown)


bsipocz (approved):
This is a very niche corner case, nevertheless thank you for fixing it!
The code diff looks all good to me.

PR #729, opened 2026-02-06
Move `astropy.samp` to `pyvo.astropy_samp`

fjebaker:
Addresses #155 (mirrored at https://github.com/astropy/astropy/issues/9763).

(should it be called `pyvo.astropy_samp`?  In my opinion it would be cleaner to merge with the existing SAMP things so it's all under `pyvo.samp`?)

Not much had changed in the astropy/main version of the SAMP implementation since #527, so I applied the diffs on top as a fixup commit.

The only additional change required was a test case that I don't understand why it wasn't failing before? From my understanding of the exception subclass rules, `SAMPProxyError` could never be caught when a `xmlrpc.client.Fault` was being thrown.

I had the tests for SAMP all passing (some non-SAMP ones were failing though) and then rebased onto the main branch of pyvo. I'm now hitting this error when I try to use pytest:

```
  File "/home/fergus/.local/share/hatch/env/virtual/pyvo/jgqKZZnG/pyvo/lib/python3.14/site-packages/pyvo/astropy_samp/constants.py", line 36, in <module>
    with open(get_pkg_data_filename("data/astropy_icon.png"), "rb") as f:
              ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fergus/.local/share/hatch/env/virtual/pyvo/jgqKZZnG/pyvo/lib/python3.14/site-packages/astropy/utils/data.py", line 770, in get_pkg_data_filename
    return download_file(
        conf.dataurl + data_name,
    ...<3 lines>...
        sources=[conf.dataurl + data_name, conf.dataurl_mirror + data_name],
    )
  File "/home/fergus/.local/share/hatch/env/virtual/pyvo/jgqKZZnG/pyvo/lib/python3.14/site-packages/astropy/utils/data.py", line 1586, in download_file
    raise urllib.error.URLError(
        f"Unable to open any source! Exceptions were {errors}"
    ) from errors[sources[0]]
urllib.error.URLError: <urlopen error Unable to open any source! Exceptions were {'http://data.astropy.org/data/astropy_icon.png': <HTTPError 404: 'Not Found'>, 'http://www.astropy.org/astropy-data/data/astropy_icon.png': <HTTPError 404: 'Not Found'>}>
```

I don't really know how to resolve it. If I curl the URLs its trying to reach it's 200 fine? Importing `pyvo` and using SAMP works too.

Help with that much appreciated!


bsipocz:
> I think you have to keep approving until they have a PR in...

I think it should be fine once he is in the team.

bsipocz:
Locally it will work fine as long as you have the datafile. Let me just push a one line to the config, which should fix it.

bsipocz:
A few gotchas/notes:
- it's highly recommended to work from a feature branch on your fork, especially for large PRs. (I mean next time)
- running pytest directly usually doesn't pick up issues with packaging; but you almost always reproduce the issues you see in CI if running the tox commands. E.g. `tox -e test`

bsipocz:
> In the meantime, I merged everything together into a pyvo.samp module. The existing samp.py only contains a few extra helper functions for sending results to SAMP clients, so there was not much work involved.

OK, that sounds all right, push it in here. I'll leave the actual review to the others as I don't really have relevant samp experience.

bsipocz:
The sphinx warnings are real, but we can wrangle the docs once the review is done and if opinions about the public API are expressed.

pyvo/samp/utils.py
```diff
@@ -0,0 +1,217 @@
...
+
+def internet_on():
+    from . import conf
+
+    if not conf.use_internet:
+        return False
+    else:
+        try:
+            urlopen("http://google.com", timeout=1.0)
+        except Exception:
```
bsipocz: It would be probably useful to clean up all these blanket exceptions (here and everywhere below); maybe it's more as a part of a follow-up though

bsipocz (commented):
Some very high level comments, I would say the most important part will be to think about what the public API should look like.

pyvo/samp/setup_package.py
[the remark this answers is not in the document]
pllim: @astrofrog probably should have deleted it in https://github.com/astropy/astropy/pull/8413 but looks like he deleted the content but not the file itself. Probably an oversight.

msdemlei (approved):
Given https://github.com/astropy/pyvo/pull/729#issuecomment-3864511969, I'm all in favour of merging this.

pyvo/samp/tests/test_web_profile.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+
+import threading
+from urllib.request import Request, urlopen
+
+import pytest
+
+from astropy.samp import SAMPHubServer, SAMPIntegratedClient, conf
+from astropy.samp.web_profile import CLIENT_ACCESS_POLICY, CROSS_DOMAIN
+from astropy.tests.helper import CI
```
bsipocz: Just swap this out with the line, here and in the other file, too.

```
CI = os.environ.get("CI", "false") == "true"
```

fjebaker:
I tried discovering which tests are causing MacOS to fail but it's a bit tedious since I don't have a Mac and can't reproduce the failure outside of the GitHub CI. So, instead I've just skipped basically all the SAMP tests on MacOS. It's not ideal but I think if it needs to be addressed it can be done so in a subsequent PR.

bsipocz (approved):
Thank you @fjebaker, this is awesome!

Would you mind going ahead and make a deprecating CI for astropy? Or let us know if not and the either me or @pllim will make one.

bsipocz:
> I tried discovering which tests are causing MacOS to fail but it's a bit tedious since I don't have a Mac and can't reproduce the failure outside of the GitHub CI. So, instead I've just skipped basically all the SAMP tests on MacOS. It's not ideal but I think if it needs to be addressed it can be done so in a subsequent PR.

That sounds all good. The issue is that the tests passing well for me locally, on osx. So I'm just as puzzled. Would you mind opening a reminder issue for turning back on these tests?

fjebaker:
@bsipocz thanks for looking at this so quick! Yes, I can put together a deprecating PR (?) for astropy. I had a quick look at how astropy does that and I gather I should decorate all of the main SAMP classes and methods with `@deprecated(...)`?

pllim:
> I don't see dates for v8.0 yet

@astrofrog , can you please fill in https://github.com/astropy/astropy/wiki/Release-Calendar ? 🙏 

I am guessing it would be Apr-May 2026, which is just before the release here. So maybe pushing back to astropy v8.1 is more realistic. Which means we cannot merge the astropy.samp deprecation PR until after v8.0.x is branched.

Issue #720, opened 2026-01-29
Exposing a timeout parameter to run_async


stvoutsin:
## Description

`run_async` doesn't expose a timeout parameter, so users can't customize it without dropping down to the lower-level job API
Is there any reason not to add an optional timeout parameter to `run_async`:

```
def run_async(self, query, *, timeout=600., delete=True, ...):
    ...
    job.wait(timeout=timeout)
    ...
```

Would you want to open a PR for this?

PR #730, opened 2026-02-09
ENH: Add timeout parameter to run_async method in tap module

stvoutsin:
## Description

This exposes the timeout parameter to the `run_async` method, allowing a user to specify a maximum runtime for an asynchronous query before it times out.

This closes issue #720

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 Enhancements and Fixes
 ----------------------
 
+- Add timeout parameter to run_async method in tap module [#730]
+
 - Fix SIA2 overflow warnings [#727]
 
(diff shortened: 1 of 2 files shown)


bsipocz (approved):
Thank you, it all looks good to me.

I would still like to see one other approval and maybe opinions about the default 600s value, as well as if we consider this API change or bugfix -- whether to release nowish with 1.8.1 or later in May with 1.9

stvoutsin:
> The 600s timeout is the same than in `wait`, so it doesn't change the previous behavior. It's true that duplicating the literal can lead to issues later though. Should we set a global variable?

Yea that seems like a good idea, I've set this as a global variable. Let me know if this looks better or if any other changes are needed to the PR.

ManonMarchand (approved):

PR #736, opened 2026-02-24
fix: preserve text/plain error body in raise_if_error() after stream is consumed

stvoutsin:
## Description

DALServiceError raised by `raise_if_error` was showing an empty message
Example:
  `DALServiceError:  for https://.../sync`

when the caller used the pattern:
```
data = sq.execute_stream().read()
sq.raise_if_error() 
```

The root cause seemed to be that `execute_stream` returns the raw response stream without reading the body. 
If a caller consumed that stream before calling `raise_if_error` the  `response_text` that `DALServiceError.from_except` reads was empty because presumably the underlying buffer was already exhausted.

## Fix

The fix reads and caches the response body in  the exception handler, before the raw stream is returned to the caller. This ensures response.text remains available when `raise_if_error` is called later regardless of whether the caller has already read the stream.

### DIFF
--- a/pyvo/dal/query.py
+++ b/pyvo/dal/query.py
@@ -211,7 +211,10 @@ class DALQuery(dict):
         try:
             response.raise_for_status()
         except requests.RequestException as ex:
-            # save for later use
+            # Cache the response body before returning the raw stream. If the
+            # caller consumes the stream before calling raise_if_error()
+            # this would lead to an empty error message
+            _ = response.content
             self._ex = ex
 
         return response.raw
--- a/pyvo/dal/query.py
+++ b/pyvo/dal/query.py
@@ -211,11 +211,7 @@ class DALQuery(dict):
         try:
             response.raise_for_status()
         except requests.RequestException as ex:
-            # Cache the response body before returning the raw stream. If the
-            # caller consumes the stream before calling raise_if_error()
-            # this would lead to an empty error message
-            _ = response.content
-            self._ex = ex
+            self._ex = DALServiceError.from_except(ex, self.queryurl)
 
         return response.raw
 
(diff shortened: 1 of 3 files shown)


pyvo/dal/query.py
```diff
@@ -211,7 +211,10 @@ def execute_stream(self, *, post=False):
         try:
             response.raise_for_status()
         except requests.RequestException as ex:
-            # save for later use
+            # Cache the response body before returning the raw stream. If the
+            # caller consumes the stream before calling raise_if_error()
+            # this would lead to an empty error message
+            _ = response.content
```
msdemlei: Hu.  That's too magic for me.  How would this local reference "cache" the response body?  I've not researched the implementation of the content attribute in requests; it probably reads the stream as a side effect?

If that's what it is, there should at least be a comment to that effect, but frankly I'd like that to be a lot more explicit.  Do we have a strong reason to not digest the respone content to a user-accessible message right here?

msdemlei (changes requested):

pyvo/dal/query.py
stvoutsin: Ok I don't disagree with that take.
The initial approach of accessing `response.content `worked because requests caches the body as a side effect, but that's implicit and probably fragile.

I've updated the fix to build the `DALServiceError` eagerly in `execute_stream` while the response body is still available, rather than deferring to `raise_if_error`. 
Now `raise_if_error` will just re-raise the constructed exception  so there should be no timing dependency on stream consumption now.

Does this seem like a better approach?

msdemlei (approved):
Yeah, that's a lot clearer now.  Thanks!

PR #743, opened 2026-03-24
FIX: raise_if_error() to use cached job phase instead of re-polling

stvoutsin:
## Description
`raise_if_error`() currently calls `self.phase`, which triggers `_update()` which causes an extra HTTP request to the job endpoint. However `raise_if_error` is called `wait()` or `run_async()` where the job has already been updated, so it makes sense to use the cached terminal job state from `self._job`

### Tests
Added `test_raise_if_error_uses_cached_phase` which verifies that calling `raise_if_error()` on an errored job does not increase the HTTP request count.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 Enhancements and Fixes
 ----------------------
 
+- Fix raise_if_error() to use cached job phase instead of re-polling [#743]
+
 - Add ``pyvo.registry.get_RegTAP_service_url()`` to allow users to inspect the
   currently configured registry endpoint. Expose other user-facing constants at their package level. [#742]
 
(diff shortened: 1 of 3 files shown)


bsipocz (approved):
Thank you!

msdemlei (approved):
I was thinking a bit about how this could go wrong.

So, when the job hasn't been started yet, it should still be fine because _update is called as part of the constructor and then _job exists.

Will this hide problems that would previously have been visible?  I don't think so, because the standard pattern is to poll the state and only after a state change call raise_if_error.  The only thing where is might break something that was previously working would be a loop like:

```
while True:
     job.raise_if_error()
     time.sleep(10)
```

Outside of test code, I think this would be insane.  And even for test code, this would be exceedingly daring.  I'd count code like this as buggy to begin with.  So, let's merge.

Issue #707, opened 2025-11-14
Regularise discovery for fragment-identified standards


msdemlei:
In the Servicetype constraint (in registry.rtcons), only “old-style” standardIDs (which are references to full registry records) are supported without hacks.  “New-style” references like SIA2's (e.g., ivo://ivoa.net/std/sia#query-2.0) require extra code, partly because the mapping is not as trivial as when we currently build SERVICE_TYPE_MAP, partly because we cannot compare with equality but need to RE-compare to ignore the minor version.

Since we now have three or the new-style standard_ids (see ``Servicetype.__init__``), let's come up with some way to keep them more declaratively, too.

PR #744, opened 2026-03-24
ENH: Declaratively define new-style standard IDs in Servicetype constraint

stvoutsin:
The Servicetype constraint handled sia2, hats, and hips through separate elif branches, raw SQL strings in extra_fragments  and a fragile string-comparison.

This adds a module-level `NEW_STYLE_STDIDS` dict mapping the standard shorthand keys to its LIKE patterns and replaces extra_fragments with a like_patterns set.

With this approach adding a new new-style standard now requires a single line in `NEW_STYLE_STDIDS`.

Does this address https://github.com/astropy/pyvo/issues/707 the way you were thinking @msdemlei ?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ Enhancements and Fixes
 
 - Support VOTableFile in accessible_table and broadcast_samp [#745]
 
+- Declaratively define new-style standard IDs in Servicetype constraint [#744]
+
 - Fix raise_if_error() to use cached job phase instead of re-polling [#743]
 
 - Add ``pyvo.registry.get_RegTAP_service_url()`` to allow users to inspect the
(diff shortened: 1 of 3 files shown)


msdemlei (approved):
This is much better than it was before, thanks.

However, now that I look I'm pretty sure the hips identifier is not what we want.  What's in there right now will find hipslists, which is not something that's useful with pyvo at this point.

The pattern should be for hipses themselves and hence: ivo://ivoa.net/std/hips#hips-1.%.

Since this probably hasn't been in use much, I'd be fine changing this with this PR and much commenting.  If you don't change it right here, let's make another PR for that right away before people become confused.

I also eye the hats identifier with suspicion.  The HATS note doesn't say anything about capability ids yet; let's take that debate to the registry list; HATS folks: you should go there anyway, because once you are writing ivo://ivoa.net/std/hats, there should something corresponding to that in the registry, and we can help you with that.

msdemlei (approved):
Thanks!

Issue #734, opened 2026-02-24
Metadata-preserving samp.send_table_to


msdemlei:
So far, I've been telling people to do  pyvo.samp.send_table_to(tap_result.to_table()).  Turns or that is a bad idea because it is losing essentially all metadata: PARAMs and INFOs are lost in this way, and that's bad not only since data origin.  For instance, it also loses the spectral axis when playing with the Heidelberg sampled XP spectra.

We should therefore tell people to send tap_result.votable.

That doesn't work so far because send_table_to uses "unified I/O", and tap_result.votable doesn't have a write method.  Actually, it's the accessible_table context manager that fails.  A simple way to enable sending VOTables as such would be:

```
@contextlib.contextmanager
def accessible_table(table):
    """
    a context manager making astropy_table available under a (file)
    URL for the controlled section.
    """
    handle, f_name = tempfile.mkstemp(suffix=".xml")
    with open(handle, "w") as f:
        if hasattr(table, "to_xml"):
            table.to_xml(f)
        else:
            table.write(output=f, format="votable")
    try:
        yield "file://" + f_name
    finally:
        os.unlink(f_name)
```

but I suspect the hasattr won't go well with... umm, anyone but me.

What would be wrong with giving VOTable objects a unified I/O-like interface, enough at least so that accessible_table would be happy with it?

Or am I missing a really good solution to this problem?  Should we perhaps explain all kinds of service results to send_table_to?

PR #745, opened 2026-03-26
FIX: Support VOTableFile in accessible_table and broadcast_samp

stvoutsin:
This PR is motivated by https://github.com/astropy/pyvo/issues/734 where metadata was lossed in the process of broadcasting TAP results via SAMP as both `broadcast_samp` and  `send_table_to` which both used `to_table()`.

The fix here is to pass the votable directly instead and update accessible_table to handle VOTableFile objects.
There is an associated astropy change here: https://github.com/astropy/astropy/pull/19499 which would add a write method to `VOTableFile` via the unified I/O framework allowing the `isinstance` fallback in `accessible_table` to be removed once that lands.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 Enhancements and Fixes
 ----------------------
 
+- Support VOTableFile in accessible_table and broadcast_samp [#745]
+
 - Fix raise_if_error() to use cached job phase instead of re-polling [#743]
 
 - Add ``pyvo.registry.get_RegTAP_service_url()`` to allow users to inspect the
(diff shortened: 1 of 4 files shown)


msdemlei (approved):
Let's have this as a straightforward fix, thanks.

It's probably not unreasonable to only special-case VOTables, in particular because that's what SAMP talks about.  However, looking forward it would be nice if we had some way to choose the most metadata-rich representation automatically (think: parquet file with extra VO metadata; one day, we might pull stuff like this from TAP, and then might not have a votable attribute) rather than the most metadata-poor (the astropy table).  I can't say I have a good plan for that, though.

bsipocz (approved):
I have one minor comment about a future cleanup; otherwise let's merge this.

Thanks!

PR #747, opened 2026-04-17
MivotInstance update

lmichel:
- FEATURE: Return None when accessing missing attributes, e.g. not mapped model field
- BUG FIX: Support unit strings with a DOT, e.g. ms.year-1

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@
 Enhancements and Fixes
 ----------------------
 
+- Improve ``pyvo.mivot.viewer.%ivotInstance`` (API to allow users to access the mapped data
+  attributes directly as properties of the instance); [#747]
+  - Access to not exiting attribute, e.g. not mappped, returns ``None``
+  - Support DOTS in unit string, e.g. "mas.year_1"
+  
 - Support VOTableFile in accessible_table and broadcast_samp [#745]
 
 - Declaratively define new-style standard IDs in Servicetype constraint [#744]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ Enhancements and Fixes
 - Improve ``pyvo.mivot.viewer.%ivotInstance`` (API to allow users to access the mapped data
   attributes directly as properties of the instance); [#747]
   - Access to not exiting attribute, e.g. not mappped, returns ``None``
-  - Support DOTS in unit string, e.g. "mas.year_1"
+  - Support DOTS in unit string, e.g. "mas.year-1"
   
 - Support VOTableFile in accessible_table and broadcast_samp [#745]
 
--- a/pyvo/mivot/tests/test_mivot_instance.py
+++ b/pyvo/mivot/tests/test_mivot_instance.py
@@ -5,6 +5,7 @@ Created on 19 févr. 2024
 @author: michel
 '''
 import pytest
+from copy import deepcopy
 from astropy.table import Table
 from pyvo.mivot.viewer.mivot_instance import MivotInstance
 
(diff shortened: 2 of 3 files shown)


msdemlei:
On Fri, Apr 17, 2026 at 09:07:55AM -0700, Laurent MICHEL wrote:
> - BUG FIX: Support unit strings with a DOT, e.g. ms.year-1

Since I just spotted it: That's not a valid VOUnits string.  As it
says in the short VOUnits summary: "Raising a unit to a power is
indicated only with the symbol ‘**’ (not a ‘^’ or by simply appending
an integer)".  But you certainly shouldn't parse VOUnits yourself
anyway.  There's a parser for them in astropy.

lmichel:
@msdemlei, 

Thank you for your comments.

- The mivot module does not parse the units, as you said this is the job of Astropy which will take place later. The purpose of this PR is just to ensure that units are read correctly regardless of their values.
- The tested units (`mas.year-1`) come from TAP simbad. Whether they are valid or not, the API must be able to read them.

I took this opportunity to simplify the code in question a bit and expand the tests.

pyvo/mivot/tests/test_mivot_instance.py
```diff
@@ -121,6 +122,36 @@ def test_mivot_instance_update():
...
+
+def test_mivot_instance_unit_filtering():
+    """Test that unit are properly read and stored in the class instance.
+    """
+    mydict = deepcopy(fake_hk_dict)
+    mydict["longitude"]["unit"] = "mas.year-1"
+    mivot_object = MivotInstance(**mydict)
+    assert mivot_object.longitude.unit == "mas.year-1"
+    mydict["dmid"] = "a.b.c"
+    mivot_object = MivotInstance(**mydict)
```
tomdonaldson, line 142: Should there be an `assert` for this object or is it just ensuring that no exceptions are thrown on the instantiation?

tomdonaldson (commented):
Thanks @lmichel ,this looks good to me.  I had one minor question, but will approve.

There is also a minor merge conflict.  I can resolve that if it's helpful.

pyvo/mivot/tests/test_mivot_instance.py
lmichel: It just to check that  no exceptions are thrown

lmichel:
Conflict in change log fixed

CHANGES.rst
```diff
@@ -3,11 +3,16 @@
...
 ----------------------
-
+  
 - Change get_tables in regtap.py to use 2 calls instead of N + 1 [#750]
 
 - Clear stale _ex before each execute_stream call [#751]
 
+- Improve ``pyvo.mivot.viewer.%ivotInstance`` (API to allow users to access the mapped data
+  attributes directly as properties of the instance); [#747]
+  - Access to not exiting attribute, e.g. not mappped, returns ``None``
```
andamian, line 13:
```suggestion
  - Access to non-existent attribute, e.g. not mapped, returns ``None``
```

CHANGES.rst
```diff
@@ -3,11 +3,16 @@
...
+  - Support DOTS in unit string, e.g. "mas.year-1"
```
andamian, line 14: Not obvious how this was achieved? By removing code?

andamian (approved):

CHANGES.rst
lmichel: by implementing a `__getattr__(self, item)` method. Should the change log give this detail?

bsipocz: nope, no need for these internal details in the changelog. If you want you can leave some comments in the source code; but I'm not sure how much clearer it would make things.

lmichel: There is a long comment telling what this method is for

tomdonaldson (approved):

PR #751, opened 2026-05-16
FIX: clear stale _ex before each execute_stream call

stvoutsin:
## Description

`_ex` was never cleared between `execute_stream` calls, so there was the potential that a stale HTTP error from a previous failed request could mask VOTable parse errors on a later successful one.

## Fix

Reset `self._ex = None` at the start of `execute_stream`


msdemlei (approved):
Thanks!

PR #752, opened 2026-05-19
docs(README): drop dead VAO website link, point at Wayback snapshot (#748)

### LINKED ISSUE #748: Bad link in README
This link at the top of the README is not valid: 

https://www.usvao.org/

It goes to a generic "buy a domain" site now.

mvanhorn:
Closes #748.

`www.usvao.org` was the original VAO project website that funded the early development of PyVO. The domain has since lapsed and now resolves to a parking page that fingerprints visitors and offers the domain for sale, which is what the issue reports.

Drop the inline URL from the parenthetical and add a one-line note pointing at the 2019-01-06 Wayback snapshot so the historical project page stays reachable for anyone tracing the package's origins. The surrounding context is preserved as-is.

### Change Type

- [x] Documentation

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,8 +21,10 @@ PyVO is a package providing access to remote data and services of the
 Virtual observatory (VO) using Python.
 
 Its development was launched by the NSF/NASA-funded Virtual Astronomical
-Observatory (VAO, www.usvao.org) project (formerly under the name
-VAOpy) as part of its initiative to bring VO capabilities to desktop.
+Observatory (VAO) project (formerly under the name VAOpy) as part of its
+initiative to bring VO capabilities to desktop. The VAO website has since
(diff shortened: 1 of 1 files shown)


msdemlei (approved):
Thanks!

mvanhorn:
Thanks for the quick turnaround! If the Wayback link isn't quite the shape you want long-term, happy to push a follow-up that either drops the parenthetical entirely or replaces it with a current IVOA/VAO reference. Just let me know.

bsipocz:
I'm sorry for not reacting to the PR sooner, but I'm afraid I'll edit out the link and wayback machine from the readme. The link already redirects to navo, and it's not exactly the same thing so I think it will be cleaner to just do a cleanup.

Issue #755, opened 2026-05-21
BUG: Copying a TAPService has side affects


emellega:
When copying a TAPService instance with an auth session we get some kind of validation as a side affect. Here's a minimal python script to reproduce the issue:

```
import pyvo, copy
auth = pyvo.auth.authsession.AuthSession()
tap = pyvo.dal.TAPService('https://sky.esa.int/esasky-tap/tap', session=auth)
copy.deepcopy(tap)
```

It produces this warning:

> WARNING: W19: ?:?:?: W19: TAP Capabilities must not have an ivo-id other than ivo://ivoa.net/std/TAP [pyvo.io.vosi.tapregext]
WARNING:astropy:W19: ?:?:?: W19: TAP Capabilities must not have an ivo-id other than ivo://ivoa.net/std/TAP

This issue was noticed when caching requests in astroquery. This [PR](https://github.com/astropy/astroquery/pull/3586) introduces a temporary workaround in astroquery.

PR #756, opened 2026-05-25
Deep copy bug in ElementWithXSIType

andamian:
Fix for #755

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,10 @@ Enhancements and Fixes
 ----------------------
 
 - Fix auth methods registered for base URLs being dropped for capability-declared sub-paths [#758]
-  
+
+- Fix W19 when deep-copying ``TAPService`` (and other objects using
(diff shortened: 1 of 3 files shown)


bsipocz (commented):
Could you please add a chaneglog entry? I don't think I'll do another bugfix, next release will be 1.9 sometimes next week.

 I also wonder if we could do a reasonable test that trips without this PR.

andamian:
> Could you please add a chaneglog entry? I don't think I'll do another bugfix, next release will be 1.9 sometimes next week.
> 
> I also wonder if we could do a reasonable test that trips without this PR.

I was hoping to pick up another issue or two but ran out of time...

bsipocz:
> I was hoping to pick up another issue or two but ran out of time...

I totally hear you.

OTOH, I'm happy to do more bugfix releases between the interops when/if we have PRs merged :)

bsipocz (approved):
Thanks!

PR #757, opened 2026-05-26
ENH: Change tapregext code to handle proposed use of forMode to allow specifying limits per mode

stvoutsin:
## Description

This is my attempt to provide client support for the proposal to add a `forMode` attribute to capabilities to allow specifying mode-specific (`sync` vs `async`) limits (`executionDuration`, `retentionPeriod`, `outputLimit`, and `uploadLimit`) as written in https://github.com/ivoa-std/TAPRegExt/pull/10.

The current behaviour is that PyvVO will warn if more than 1 of these elements appears, which makes sense for the existing TAPRegExt spec.

Since this is still a PR in `TAPRegExt` I'll leave this as a Draft PR mainly for visibility and discussion. If we don't want this cluttering the PR list I could also move this into an issue and add a comment to refer to the branch with the changes.

## Changes

Limit elements (`executionDuration`, `outputLimit` etc.) are now parsed into lists with new `get_x(mode)` methods for mode-aware getters. 

One place I noticed that can take advantage of these mode-specific limits is the `run_async` timeout, so I've changed that to default to the service's advertised async execution duration as its wait timeout.

## Compatibility

With the approach I've gone with existing code that depends on the singular limit properties (maxrec, and hardlimit) should be fine.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,9 @@ Deprecations and Removals
 Enhancements and Fixes
 ----------------------
 
+- Add support for the forMode attribute in TAPRegExt, which allows TAP services 
+  to advertise different limits for different modes [#757]
+  
 - Fix auth methods registered for base URLs being dropped for capability-declared
   sub-paths [#758]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,9 +27,14 @@ Deprecations and Removals
 Enhancements and Fixes
 ----------------------
 
-- Add support for the forMode attribute in TAPRegExt, which allows TAP services 
+- Add support for the forMode attribute in TAPRegExt, which allows TAP services
   to advertise different limits for different modes [#757]
-  
+
+- Change the default ``timeout`` behavior of ``TAPService.run_async``. When
+  ``timeout`` is not specified, it now uses the advertised async
+  ``executionDuration`` default instead of falling back to
+  ``DEFAULT_JOB_WAIT_TIMEOUT``. [#757]
+
 - Fix auth methods registered for base URLs being dropped for capability-declared
   sub-paths [#758]
 
--- a/pyvo/dal/tap.py
+++ b/pyvo/dal/tap.py
@@ -255,6 +255,47 @@ class TAPService(DALService, AvailabilityMixin, CapabilityMixin):
             pass
         raise DALServiceError("Hard limit not exposed by the service")
 
+    def get_maxrec(self, mode=None):
+        """
+        The default output limit for the given mode.
+
+        Parameters
+        ----------
+        mode : str or None
+            The access mode ('sync', 'async', or None for the default limit).
+
+        Raises
+        ------
+        DALServiceError
+            if the property is not exposed by the service
+        """
+        try:
+            return self.get_tap_capability().get_outputlimit(
+                mode).default.content
+        except AttributeError:
+            pass
+        raise DALServiceError("Default limit not exposed by the service")
+
+    def get_hardlimit(self, mode=None):
+        """
+        The hard output limit for the given mode.
+
+        Parameters
+        ----------
+        mode : str or None
+            The access mode ('sync', 'async', or None for the default limit).
+
+        Raises
+        ------
+        DALServiceError
+            if the property is not exposed by the service
+        """
+        try:
+            return self.get_tap_capability().get_outputlimit(mode).hard.content
+        except AttributeError:
+            pass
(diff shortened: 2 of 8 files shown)


msdemlei:
On Tue, May 26, 2026 at 10:55:37AM -0700, stvoutsin wrote:
> Since this is still a PR in `TAPRegExt` I'll leave this as a Draft

Oh dang, sorry I've let this slip.

I've closed ivoa-std/TAPRegExt#8, and I'm pretty sure
ivoa-std/TAPRegExt#10 is basically fine for merging.  Would someone
review it?

Once it's in, I'd say we should just merge this.  Ok, it's not in a
REC yet, but then I'm rather sure nothing much will break if we do
minor adjustments later, and we should really get the ball rolling
properly.

stvoutsin:
> On Tue, May 26, 2026 at 10:55:37AM -0700, stvoutsin wrote: Since this is still a PR in `TAPRegExt` I'll leave this as a Draft
> Oh dang, sorry I've let this slip. I've closed [ivoa-std/TAPRegExt#8](https://github.com/ivoa-std/TAPRegExt/pull/8), and I'm pretty sure [ivoa-std/TAPRegExt#10](https://github.com/ivoa-std/TAPRegExt/pull/10) is basically fine for merging. Would someone review it? Once it's in, I'd say we should just merge this. Ok, it's not in a REC yet, but then I'm rather sure nothing much will break if we do minor adjustments later, and we should really get the ball rolling properly.

Oh no problem, I had also forgotten about this one until now. I've added a couple of very minor comments to that TAPRegExt PR but as far as I'm concerned it looks ready to merge.

msdemlei (approved):
I'd say we can merge this (and at least I know the author:-).

As a matter of style and parsimony, I'd prefer if the various test capability documents could be templated, such that the varying parts are being inserted in the fixture; I'm thinking of something like

```
@pytest.fixture
def job_with_duplicate_formode(...):
  with open(get_pkg_data_filename(...template..., "rb") as f:
    return io.BytesIO(f.read().format("""
    <executionDuration forMode="sync">
      <default>60</default>
    </executionDuration>
    <executionDuration forMode="async">
      <default>7200</default>
    </executionDuration>"""))
```

– but that preference isn't strong enough that I'd do that myself.

And then the question is whether we want code in pyVO that's just implementing a WD.

I'd give that a vigouros yes for so many reasons I'd be happy to discuss in a separate bug.

stvoutsin:
> I'd say we can merge this (and at least I know the author:-).
> 
> As a matter of style and parsimony, I'd prefer if the various test capability documents could be templated, such that the varying parts are being inserted in the fixture; I'm thinking of something like
> 
> ```
> @pytest.fixture
> def job_with_duplicate_formode(...):
>   with open(get_pkg_data_filename(...template..., "rb") as f:
>     return io.BytesIO(f.read().format("""
>     <executionDuration forMode="sync">
>       <default>60</default>
>     </executionDuration>
>     <executionDuration forMode="async">
>       <default>7200</default>
>     </executionDuration>"""))
> ```

Happy to make that change, so I'll update this PR hopefully sometime next week, or as a follow up PR if this gets merged in before then.

stvoutsin:
> As a matter of style and parsimony, I'd prefer if the various test capability documents could be templated, such that the varying parts are being inserted in the fixture; I'm thinking of something like
> 
> ```
> @pytest.fixture
> def job_with_duplicate_formode(...):
>   with open(get_pkg_data_filename(...template..., "rb") as f:
>     return io.BytesIO(f.read().format("""
>     <executionDuration forMode="sync">
>       <default>60</default>
>     </executionDuration>
>     <executionDuration forMode="async">
>       <default>7200</default>
>     </executionDuration>"""))
> ```
> 

I've changed the tests to be templated and in the process made a small change to `test_rtcons.py` to address pytest failures due to a pytest 9.1 upgrade which deprecates the `scope="class"` pattern (now changed to a class method)

msdemlei (approved):
Looks good to me.  Perhaps someone else can have another look, and then let's merge.

CHANGES.rst
```diff
@@ -27,6 +27,9 @@ Deprecations and Removals
 Enhancements and Fixes
 ----------------------
 
+- Add support for the forMode attribute in TAPRegExt, which allows TAP services 
+  to advertise different limits for different modes [#757]
```
ManonMarchand, line 31: there should also be a mention of the change of behavior with the timeout

ManonMarchand (approved):

bsipocz:
With two approvals, let's merge this now. Thanks @stvoutsin!

PR #758, opened 2026-05-28
FIX: Auth methods registered for base URLs being dropped for capability-declared sub-paths

### LINKED ISSUE #749: BUG: allowed_auth_methods drops caller-registered auth when capabilities declare more-specific sub-path URLs
## Description

So we've run into an issue related to authentication for our services (TAP / SIA) where we use bearer tokens and add these to an AuthSession.
Basically when a caller registers an authentication method for a base URL (e.g. /tap) and pyvo then loads VOSI capabilities that declare a security method for a more-specific URL (`/tap/tables`), the caller's method is silently ignored for requests to that sub-path.
The result is that the auth negotiation failure even though the caller explicitly provided valid credentials.

## Steps to reproduce

(Assumes we have /capabilities that declare the VOSI capability with a set of IVOA-accepted auth mechanisms)
```
from pyvo.auth.authsession import AuthSession
from pyvo.utils.http import create_session
import pyvo.dal

session = AuthSession()
credentials = create_session()
credentials.headers["Authorization"] = "Bearer mytoken"
session.credentials.set("my-token", credentials)
session.add_security_method_for_url("https://example.com/tap", "my-token")

service = pyvo.dal.TAPService("https://example.com/tap", session=session)
service.tables
```

## Cause

`service.tables` triggers a request to `/tap/tables`, pyvo first fetches `/tap/capabilities`, which declares a security method (e.g. ivo://ivoa.net/sso#cookie) for `/tap/tables`. 
`AuthURLs.allowed_auth_methods` then returns only that IVOA method for `/tap/tables` requests, ignoring the my-token registration on the parent /tap. 
Since the client has no cookie credentials, negotiation fails and raises an exception.

This is problematic for services like ours that use a non-IVOA auth scheme (bearer tokens) and register it via `add_security_method_for_url` on a base url, as we silently lose that authentication mechanism for sub-paths silently. As a workaround we can and have until now registered each possible subpath via `add_security_method_for_url` but this isn't ideal.

## Possible solutions

A) Union all matching `base_urls` entries (The simplest fix)
Change `allowed_auth_methods` to accumulate methods from all matching prefix entries instead of stopping at the longest.
Fix is probably the simplest, and fixes this bug. The downside is however that it mixes caller registrations and capability entries in one dictionary, so this would cause issues if for example someone wants an anonymous endpoint declared in the capabilities., which may lead to us sending authentication headers to it.

B) Special case "most-specific match is anonymous"
Basically apply A) but if the most specific matching entry is only anonymous, return and skip the union. Reason being that if a service explicitly marks an endpoint anonymous, we don't silently want to send credentials to it. Downside is that this conflates anonymous is allowed with anonymous is required and I don't know how we handle the case where a service lists both anonymous and a real method.

C) Separate caller registrations from capability-sourced entries (My preference)
With this solution we'd have two distinct dicts:
- `_prefix_urls` (Caller registrations)
- `_capability_urls` (Capability-sourced entries)
`_update_from_capabilities` would be a new method that writes to `_capability_urls`, and then `add_security_method_for_url` would write to `_prefix_urls`. `allowed_auth_methods` would be the union of all matching `_prefix_urls` entries and the single most-specific `_capability_urls` entry, after checking `full_urls` (exact match) first.

I think this is probably the cleaner long-term fix, since it encodes the intent of the registration of authentication mechanisms at the different points, rather than relying on some obscure lookup logic. `full_urls` in capabilities would continue to act as an override, which correctly handles services that want to strictly restrict auth on a specific endpoint. This is unfortunately also the more complicated and a more involved set of changes to the internal APIs.

Sorry for the ramble, any thoughts on this? I'm happy to work on implementing whatever approach folks think is the best here. (Assuming people agree that this needs fixing)

stvoutsin:
## Description

This PR is my attempt to address issue #749 where non-IVOA recognized authentication methods added by the users via `add_security_method_for_url` (e.g. bearer tokens) are replaced by auth methods declared in the VOSI capabilities for more specific sub-paths (/tap/tables). Auth fails in this case as the client's custom credentials are no longer in the list of available methods.

## Cause

`AuthURLs` stored both caller registrations and capability-discovered methods in the same `base_urls` dict, then used longest-match to pick one entry. 
A capability entry for `/tap/tables` would win over the caller's /tap registration, discarding it entirely.

## Fix

The fix proposed here is to split the `base_urls` into two collections:
- `_explicit_urls` : Caller registrations via `add_security_method_for_url`. Methods registered for a base URL apply to all URLs beneath it.
-  `_capability_urls`: Populated from VOSI capabilities, only the most-specific (longest) matching entry is used

## Tests
Added a few tests covering the various related use-cases

Fixes: https://github.com/astropy/pyvo/issues/749

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@
 
 Enhancements and Fixes
 ----------------------
+
+- Fix auth methods registered for base URLs being dropped for capability-declared sub-paths [#758]
   
 - Change get_tables in regtap.py to use 2 calls instead of N + 1 [#750]
 
--- a/pyvo/auth/authurls.py
+++ b/pyvo/auth/authurls.py
@@ -11,11 +11,36 @@ class AuthURLs():
     AuthURLs helps determine which security method should be used
     with a given URL.  It learns the security methods through the
     VOSI capabilities, which are passed in via update_from_capabilities.
+
+    Three collections are used internally:
+
+    ``full_urls``
+        Exact-match entries, populated when a capability declares
+        ``use="full"``. An exact match takes priority and is
+        returned without consulting any other collection.
+
+    ``_explicit_urls``
+        Prefix-match entries registered by callers via
+        ``add_security_method_for_url``. All matching entries are
+        combined, so a registration for a base URL propagates
+        to every sub-path beneath it.
+
+    ``_capability_urls``
+        Prefix-match entries loaded from VOSI capabilities
+        (``use="base"``).  Only the most-specific (longest) matching
+        entry is used.
+
+    For a given URL ``allowed_auth_methods`` returns the union of:
+
+    1. The ``full_urls`` exact match, if one exists (short-circuits).
+    2. All matching ``_explicit_urls`` entries.
+    3. The single most-specific matching ``_capability_urls`` entry.
     """
 
     def __init__(self):
         self.full_urls = collections.defaultdict(set)
-        self.base_urls = collections.defaultdict(set)
+        self._explicit_urls = collections.defaultdict(set)
+        self._capability_urls = collections.defaultdict(set)
 
     def update_from_capabilities(self, capabilities):
         """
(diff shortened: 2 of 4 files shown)


andamian (approved):
This looks good to me and implements the solution discussed in the issue.

bsipocz:
Thank you for the PR and the review!

PR #760, opened 2026-06-04
FIX: Corrections to the TAPResults documentation

stvoutsin:
I think the doc string for the `TAPResults` method was copied over so it still refers to SIA query information.
PR updates the documentation accordingly.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 Enhancements and Fixes
 ----------------------
 
+- Corrections to the TAPResults documentation [#760]
+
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,8 @@
 Enhancements and Fixes
 ----------------------
 
+- Corrections to the TAPResults documentation [#760]
```
bsipocz, line 7: no need for changelog for docs and internals changes. I'll clean this up before release.

bsipocz (approved):
Thanks!

Issue #761, opened 2026-06-06
pyvo 1.9.0 attempts to overwrite samp_hub


Schamschula:
As I was updating the MacPorts Portfile for `py-pyvo` to version 1.9.0 I got the following error:

```
--->  Activating py314-pyvo @1.9.0_0
Error: Failed to activate py314-pyvo: The following ports have active files that conflict with py314-pyvo's:
py314-astropy @7.2.0_1+all
  /opt/local/Library/Frameworks/Python.framework/Versions/3.14/bin/samp_hub
  /opt/local/bin/samp_hub-3.14
```

Apparently, `samp_hub` is a duplicate file name to what `py314-astropy @7.2.0_1` has already installed at the same location.

Note: `samp_hub-3.14` is a symlink to the file installed under the Frameworks tree.

I didn't see this error for earlier versions of `py-pyvo`.

bsipocz:
Yes, this is not totally surprising. We are in the middle of moving astropy.samp downstream into pyvo and then deprecate and remove it in astropy. I label this as a big even though I don't think it's an actual bug here, but something we need to fix in astropy.

cc @astrofrog and @pllim from the astropy side to see how we should patch this problem around

astrofrog:
Should PyVO maybe use a different name like samp-hub? Or am I misunderstanding the problem?

Schamschula:
Currently my only option is to rename or remove it. My guess is that renaming does little for anyone using this binary (they will use the one in py-astropy), so I will remove it.

However, I must be careful to check updates to py-astropy to see when it is removed there.

bsipocz:
I would go the other way around, and favour this one as opposed to the one in astropy; that should have got the deprecation but we couldn't add that until pyvo was release.

PR #762, opened 2026-06-08
Fixes for SAMP migration from astropy to pyvo

pllim:
This is a follow-up of https://github.com/astropy/pyvo/pull/729

Related issue: https://github.com/astropy/pyvo/issues/761

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -95,7 +95,7 @@ pyvo.samp = data/astropy_icon.png, data/*xml
 
 [options.entry_points]
 console_scripts =
-    samp_hub = astropy.samp.hub_script:hub_script
+    samp_hub = pyvo.samp.hub_script:hub_script
 
 [coverage:run]
 source = pyvo
--- a/docs/samp.rst
+++ b/docs/samp.rst
@@ -8,10 +8,6 @@ SAMP, the Simple Application Messaging Protocol, lets you send and
 receive tables, datasets, or points of interest between various clients
 on a desktop.
 
-While the main SAMP support still resides in astropy (it might be moved
-over to pyvo in later versions), there are a few wrapper functions in
-pyvo that make a few common SAMP tasks simpler or more robust.
-
 Most importantly, pyvo lets you manage the SAMP connection in a context
 manager, which means you will not have to remember to close your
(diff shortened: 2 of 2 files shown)


pllim:
I will add a change log if you think we need it.

@astrofrog also suggested that pyvo removes the samp_hub CLI completely, so if you want to go with that instead, then this PR needs to be modified.

astrofrog:
Just to expand on my suggestion, we could move to a model where we no longer have a CLI command but set it up so that users can instead do e.g.:

```
python -m pyvo.samp.hub
```

setup.cfg
```diff
@@ -95,7 +95,7 @@ pyvo.samp = data/astropy_icon.png, data/*xml
 
 [options.entry_points]
 console_scripts =
-    samp_hub = astropy.samp.hub_script:hub_script
+    samp_hub = pyvo.samp.hub_script:hub_script
```
bsipocz, line 98: 🤦‍♀️ of not noticing this in review.

bsipocz (approved):
Thank you @pllim. I'll see to the changelog before doing a release.

bsipocz:
@tomdonaldson @fjebaker - what do you think about the cli suggestion above?

fjebaker:
I think it's a good idea! I think it fits in naturally with other Python module tools like `python3 -m http.server`.

P.S.: I am still planning to submit some SAMP patches, but unfortunately after my migration PR got merged I've struggled to find the time to work on them... :(

bsipocz:
> P.S.: I am still planning to submit some SAMP patches, but unfortunately after my migration PR got merged I've struggled to find the time to work on them... :(

No worries, you get to it when you get to it, any attention to samp on any timescale will be better than we had for it in the past many years

bsipocz:
> I think it's a good idea! I think it fits in naturally with other Python module tools like python3 -m http.server.

OK. If the other maintainers are OK with it, then I can tag a new 1.9.1 release once this change is done.

setup.cfg
pllim: It happens. No worries.

pllim:
I opened https://github.com/astropy/pyvo/pull/763

PR #763, opened 2026-06-09
API: Remove samp_hub CLI

### LINKED ISSUE #761: pyvo 1.9.0 attempts to overwrite samp_hub

pllim:
Close #761 

Removing as discussed in #762 

This is a breaking change. Also, I cannot find `samp_hub` CLI documented anywhere that I can find here, so not sure where to recommend the `python -m pyvo.samp.hub` usage that @astrofrog suggested.


bsipocz (approved):
I would say mention it in the changelog, I think we can ship this in a bugfix release as it was an oversight (and original release was done <1week, so it should all be fine); but if you think otherwise I can do a post release instead.

pllim:
You know pyvo users much better than I do, so I trust your judgement on where to put this patch. I added a change log. Thanks!

bsipocz:
@astropy/pyvo-maintainers - What do you think about the release? We have this PR and #762 as content.

 The options are, I think:

- Have a 1.9.1 release with these two PRs, do it as a regular bugfix release
- Have 1.10 as the removal of this cli is technically API change
- Have 1.9.1 as a bugfix release and yank 1.9.0 from pypi -- so this API change is removed from the users
- something else

I would go with the first option, I think it's good enough and a pragmatic solution, and no one I talked to about this has actually ever used the cli to begin with.

PR #767, opened 2026-06-13
feat: add RegistryResource.validate() for on-demand test query execution

### LINKED ISSUE #766: ENH: Add ability to run a resource's test-query
Most registry entries have some form of `test_query`, that is used for validation. 

It would be nice to have this exposed to `pyvo` users for an on-demand resource validation.

In addition to finding out if this passes or fails, it would be great to also indicate how long the test took, so multiple resources could be compared to find the most suitable resource for your network or geo-location.

mvanhorn:
## Summary

This PR adds `RegistryResource.validate()`, a new method that executes a service's registered `testQuery` on demand and reports whether the service is reachable and how long it took to respond.

## Why this matters

Most IVOA registry entries publish a `testQuery` - a known-working example request intended for validation. Today pyvo has no way to invoke it, so users who want to check service liveness or compare latency across candidate services must hand-roll their own HTTP calls. This is especially useful for HAT workflows where several equivalent services may be available and the best one depends on network proximity. The maintainer (bsipocz, MEMBER) confirmed this direction on 2026-06-11 as part of broader pyvo support for HATs.

## How it works

`validate()` follows the same remote-query pattern as `get_contact()`: it calls `get_RegTAP_service().run_sync()` to fetch the resource's testQuery parameters from `rr.res_detail` (joined to `rr.interface` to get the access URL), assembles the full request URL, performs an HTTP GET timed with `time.monotonic()`, and returns a `ValidateResult` namedtuple with `success`, `response_time_seconds`, and `status_code`. If no testQuery is registered, a `DALQueryError` is raised with an informative message. The `ValidateResult` type is exported in `__all__` for discoverability. Four unit tests cover the happy path, the no-testQuery case, an HTTP error response, and the return type of `response_time_seconds`.

Fixes #766

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@
 Enhancements and Fixes
 ----------------------
 
+- Add ``RegistryResource.validate()`` to execute a resource's registered
+  test query on demand and return a ``ValidateResult`` with ``success``,
+  ``response_time_seconds``, and ``status_code`` fields. [#766]
+
 
 Deprecations and Removals
 -------------------------
--- a/pyvo/registry/regtap.py
+++ b/pyvo/registry/regtap.py
@@ -21,6 +21,7 @@ import functools
 import itertools
 import os
 import textwrap
+import time
 import warnings
 
 from astropy import table
@@ -29,6 +30,8 @@ from astropy.utils.exceptions import AstropyDeprecationWarning
 
 import numpy
 
+import requests
+
 from . import rtcons
 from ..dal import scs, sia, sia2, ssa, sla, tap, query as dalq
 from ..io.vosi import vodataservice
@@ -36,7 +39,10 @@ from ..utils.formatting import para_format_desc
 
 
 __all__ = ["search", "get_RegTAP_query", "get_RegTAP_service_url", "Interface",
-           "RegistryResource", "RegistryResults", "ivoid2service"]
+           "RegistryResource", "RegistryResults", "ivoid2service", "ValidateResult"]
+
+ValidateResult = collections.namedtuple(
+    "ValidateResult", ["success", "response_time_seconds", "status_code"])
 
 REGISTRY_BASEURL = os.environ.get("IVOA_REGISTRY", "http://reg.g-vo.org/tap"
                                   ).rstrip("/")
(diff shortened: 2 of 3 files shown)


msdemlei (approved):
I very frankly remain unconvinced that this kind of thing is an advantage over telling science users to just do their thing and get "real" error messages if things are broken.  But then I'll not try to argue hard.  If you find something like this useful, who am I to try and talk you out of it.

I was also skeptical because the test query details are optional in RegTAP.  I've checked: All three implementations expose them.  So, I think in the next RegTAP we'll just make them mandatory on grounds that this code wants them and that they're there anyway.

I'll approve, but I think it would be great if ValidateResult could also expose the payload.  The reason there is that testQuery was really provided for the benefit of validators; these of course are interested in whether the service works at all, but the main thing is that they want to examine the query result for standards compliance (simple example: "there's a meta.id;meta.main column" in SCS).  It would be great if future validators could build upon this infrastructure, and then they'll need to look at what came back from the service.  Oh, and the response headers would be useful for that, too.

ManonMarchand:
I'd be extra careful reviewing this, it looks fully AI generated.

msdemlei:
On Mon, Jun 15, 2026 at 07:52:52AM -0700, Manon Marchand wrote:
> ManonMarchand left a comment (astropy/pyvo#767)
>
> I'd be extra careful reviewing this, it looks fully AI generated.

Thanks for the heads-up, and it's true that this could use quite a
bit more love; the idiomatic RegTAP query would be

SELECT detail_xpath, detail_value, access_url
            FROM rr.res_detail
            NATURAL JOIN rr.interface
            WHERE rd.ivoid = {}
              AND rd.detail_xpath LIKE '/capability/testQuery/%'
              AND i.intf_role = 'std'

Furthermore, we should, at some point, also evaluate the testQuery on
vs:ParamHTTP interfaces; but as I said: If this serves a purpose for
the HATS crowd, I see no strong reason to block it, either.

I'd still leave it to @bsipocz to make a call, as she is, I think, in
contact with the submitter.

bsipocz:
I would close this and not waste our time with recieving generatic ai contributions. If you would think its helpful we can merge in an AI policy that says the same that we need a human in the loop 

(The issue for this hasn't even converted on its discussion, and the whole idea started with the need for testing and example in the hats registry entries.)

msdemlei:
Fair enough.  The thing had fooled me into believing this was something the HATS folks had made.  Well, modern times.  Somewhat impressive, though.

Do we understand why someone is bothering to do this?

bsipocz:
> Do we understand why someone is bothering to do this?

honestly, I don't even think it's a someone, but a bot going through hundreds if not thousands of repos.

ManonMarchand:
Is there a mechanism to ban this bot? It targeted us twice now

bsipocz:
I think you need to be an org owner for that? Summoning @pllim.

pllim:
Blocked as requested.

PR #772, opened 2026-07-01
Pass session when creating `VOSITables` objects

fajpunk:
Session info is not propagated to `VOSITables` HTTP calls, causing auth errors with some configurations. Here's an example of this with bearer token auth. This is running against a
[youcat](https://github.com/opencadc/tap/tree/main/youcat) server instance, but the particular TAP server doesn't matter.

```python
def get_pyvo_auth(token: str, url: str) -> pyvo.auth.authsession.AuthSession | None:
    """Create a PyVO-compatible auth object."""
    s = requests.Session()
    s.headers["Authorization"] = f"Bearer {token}"

    auth = pyvo.auth.authsession.AuthSession()
    auth.credentials.set("lsst-token", s)

    for subpath in (
        "/sync",
        "/async",
        "/tables",
    ):
        auth.add_security_method_for_url(url + subpath, "lsst-token")

    return auth

auth = get_pyvo_auth(token=TOKEN, url=URL)
tap_service = pyvo.dal.TAPService(URL, session=auth)
tables = list(tap_service.tables)

# pyvo.dal.exceptions.DALServiceError: 401 Client Error: Unauthorized for url: https://my-tap-server/tables/dfuchs_test_schema.some_table
```

When accessing the `tables` property on a `TAPService` object, the existing session is not passed to to the `VOSITables` object constructor. Any auth properties of that session are not used in any HTTP requests that the `VOSITables` object makes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,13 @@ Deprecations and Removals
 1.9.2 (unreleased)
 ==================
 
-- no changes yet
+Enhancements and Fixes
+----------------------
+
+- Fix a bug where a custom session (including authentication configuration) was
+  not propagated when retrieving tables with the ``TAPService.tables``
+  property. [#772]
+
 
 1.9.1 (2026-06-11)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,11 @@ Enhancements and Fixes
   not propagated when retrieving tables with the ``TAPService.tables``
   property. [#772]
 
+Deprecations and Removals
+-------------------------
+
+  - Remove ``dal.vosi.TablesMixin``. It is no longer used anywhere. It creates a ``VOSITables`` object without propagating the existing session, and there is no clear way to fix it. It was added in https://github.com/astropy/pyvo/commit/72005872b8148fd3fee8571b357d0cbcf089d1de and all usages of it were removed in https://github.com/astropy/pyvo/commit/e71a8e3f08ef0300cd68316023bc09b186ac4ea0 on Jan 9, 2018.
+
 
 1.9.1 (2026-06-11)
 ==================
--- a/pyvo/dal/tap.py
+++ b/pyvo/dal/tap.py
@@ -177,7 +177,7 @@ class TAPService(DALService, AvailabilityMixin, CapabilityMixin):
             response.raw.read = partial(response.raw.read, decode_content=True)
 
             self._tables = VOSITables(
-                vosi.parse_tables(response.raw.read), tables_url)
+                vosi.parse_tables(response.raw.read), tables_url, session=self._session)
         return self._tables
 
     def _parse_examples(self, examples_uri, *, depth=0):
--- a/pyvo/dal/tests/test_tap.py
+++ b/pyvo/dal/tests/test_tap.py
@@ -395,33 +395,53 @@ def async_fixture_with_timeout(mocker):
     yield from mock_server.use(mocker)
 
 
-@pytest.fixture()
-def tables(mocker):
-    def callback_tables(request, context):
-        return get_pkg_data_contents('data/tap/tables.xml')
+def make_tables_fixture(custom_auth=False):
+    """
+    Returns and registers a pytest fixture to expect tables HTTP calls.
 
-    def callback_table1(request, context):
-        return get_pkg_data_contents('data/tap/lazy-table1.xml')
+    If custom_auth is true, then an auth header will be expected in all calls.
+    """
+    kwargs = {}
+    if custom_auth:
+        kwargs['request_headers'] = {'Authorization': 'Bearer some-token'}
 
-    def callback_table2(request, context):
-        return get_pkg_data_contents('data/tap/lazy-table2.xml')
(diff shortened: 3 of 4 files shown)


ManonMarchand:
LGTM, thanks! Could you add an entry in `CHANGES.rst`?

Edit: and you've got a tiny fix to do to make the linter happy. Note that you can test this locally with `tox -e codestyle`

ManonMarchand (approved):

fajpunk:
Thank you for looking at this! I found one other place where a `VOSITables` object is created:
https://github.com/astropy/pyvo/blob/78386ef943adc48662393cff8bd4d04b3355f8c7/pyvo/dal/vosi.py#L216

No session is propagated here either. It is created as part of the `TablesMixin`, but it doesn't seem like the `TablesMixin` is used anywhere. I'm not sure what to do here. Three possibilities I can think of:

* Get rid of the `TablesMixin` class. It's not used anywhere else in the PyVO codebase, and it's not in the `__all__` list of the `pybo.dal.vosi` module, but I'm not sure if any users are using it.
* Assume that any parent class using the mixin will have a `_session` attribute if a user wants to propagate a session, and dynamically check for that attribute on `self` and pass it if it exists. This seems too magical and surprising.
* Drop the `tables` attribute from `TablesMixin`, and just have a method that takes a session as a parameter. Then parents of the mixin can pass the session however it makes sense. This would also be a breaking change.

I don't have context on how `TablesMixin` is used. If there is a better way to deal with this, I'm glad to try it if you can point me in the right direction.

stvoutsin:
> Thank you for looking at this! I found one other place where a `VOSITables` object is created:
> 
> https://github.com/astropy/pyvo/blob/78386ef943adc48662393cff8bd4d04b3355f8c7/pyvo/dal/vosi.py#L216
> 
> No session is propagated here either. It is created as part of the `TablesMixin`, but it doesn't seem like the `TablesMixin` is used anywhere. I'm not sure what to do here. Three possibilities I can think of:
> 
> * Get rid of the `TablesMixin` class. It's not used anywhere else in the PyVO codebase, and it's not in the `__all__` list of the `pybo.dal.vosi` module, but I'm not sure if any users are using it.
> * Assume that any parent class using the mixin will have a `_session` attribute if a user wants to propagate a session, and dynamically check for that attribute on `self` and pass it if it exists. This seems too magical and surprising.
> * Drop the `tables` attribute from `TablesMixin`, and just have a method that takes a session as a parameter. Then parents of the mixin can pass the session however it makes sense. This would also be a breaking change.
> 
> I don't have context on how `TablesMixin` is used. If there is a better way to deal with this, I'm glad to try it if you can point me in the right direction.

I can also confirm it's not used anywhere, and further even has a bug in case it was ever used (`endpoint_url` not passed in to the `VOSITables` instantiation) so I would be in favor of just removing it, either here or maybe as a separate PR.

msdemlei:
On Thu, Jul 02, 2026 at 09:53:20AM -0700, stvoutsin wrote:
> I can also confirm it's not used anywhere, and further even has a
> bug in case it was ever used (`endpoint_url` not passed in to the
> `VOSITables` instantiation) so I would be in favor of just removing
> it, either here or maybe as a separate PR.

When we believe it's not used (and if it's broken, that is a strong
indication), I'd all be in favour of dropping it here (in a separate
commit), noting that it was broken and would have needed fixing as
part of the PR.

fajpunk:
I've removed it in another commit as @msdemlei suggested. It was added in https://github.com/astropy/pyvo/commit/72005872b8148fd3fee8571b357d0cbcf089d1de and all usages of it were removed in https://github.com/astropy/pyvo/commit/e71a8e3f08ef0300cd68316023bc09b186ac4ea0 on Jan 9, 2018.

CHANGES.rst
```diff
@@ -11,7 +11,18 @@ Deprecations and Removals
...
+----------------------
+
+- Fix a bug where a custom session (including authentication configuration) was
+  not propagated when retrieving tables with the ``TAPService.tables``
+  property. [#772]
+
+Deprecations and Removals
+-------------------------
+
+  - Remove ``dal.vosi.TablesMixin``. It is no longer used anywhere. It creates a ``VOSITables`` object without propagating the existing session, and there is no clear way to fix it. It was added in https://github.com/astropy/pyvo/commit/72005872b8148fd3fee8571b357d0cbcf089d1de and all usages of it were removed in https://github.com/astropy/pyvo/commit/e71a8e3f08ef0300cd68316023bc09b186ac4ea0 on Jan 9, 2018.
```
bsipocz, line 24: I'll clean this up before release; but this entry is contractionary:

- we do not remove (or even deprecate) things in bugfix releases
- this is all internal and is unused, so can be in fact removed. But then it doesn't in fact need a changelog entry. Either way, no need to mention commits in the changelog all that info can be traced in the PR description and comments, that is the reason we make the PR number included in the changelog entry.

fajpunk: That all makes sense, sorry for the hassle!

bsipocz: No worries, this is always great to see cleanup PRs.

PR #774, opened 2026-07-08
Use consistent column names in user table load docs

fajpunk:
The `create_table` docs create a table with a `count_of` column (because `count` is a Postgres reserved word). The `load_table` docs specify a `count` column though.

This chances `count` -> `count_of` in the data in the `load_table` docs.

### DIFF
--- a/docs/dal/index.rst
+++ b/docs/dal/index.rst
@@ -442,7 +442,7 @@ tab-separated values (tsv), comma-separated values (cvs) or VOTable (VOTable):
 .. doctest-skip::
 
     >>> tap_service.load_table(name='test_schema.test_table',
-    ...                        source=StringIO('article,count\narticle1,10\narticle2,20\n'), format='csv')
+    ...                        source=StringIO('article,count_of\narticle1,10\narticle2,20\n'), format='csv')
 
 Users can also create indexes on single columns:
 .. doctest-skip::


bsipocz:
I'm not sure I follow, we have `count` in all the xml sources, both for `create_table` and `load_table`, if anything this PR makes it inconsistent, even within the documentation.

As all of these are skipped being tested I accept if there are some actual issues with it, but then it has to be changed everywhere to be self consistent.

fajpunk:
Yes, sorry! I got two of my local branches mixed up and I mistakenly thought that `count_of` was in the `create_table` call . I am experimenting with deploying Youcat on the Rubin science platform, and I'm trying to test it with PyVO. Right now I have youcat deployed against a Postgres database, and the table definition used in the `create_table` call with the table definition in these docs errors with "Identifier 'count' is a reserved keyword."

It seems like Youcat should be responsible for whatever quoting/escaping needs to happen when creating the table, so ideally the docs here wouldn't need to change at all. I'll report this to the Youcat folks.

bsipocz:
> It seems like Youcat should be responsible for whatever quoting/escaping needs to happen when creating the table, so ideally the docs here wouldn't need to change at all. I'll report this to the Youcat folks.

We don't test that, and I don't have an account to manually test it, so if there are problems with it, or if you get back any feedback from youcat, please make a PR that updates everything.

fajpunk:
Submitted another PR to update `count` to `count_of` in both places based on feedback from the CADC folks in https://github.com/astropy/pyvo/pull/775

PR #784, opened 2026-09-14
Fix AsyncTAPJob.wait to treat timeout as a total budget not a per-request one

stvoutsin:
`wait(timeout=...)` was getting passed straight through as the per-request socket read timeout, so it never actually meant what the docstring claimed.

- Timeout was per-request, not total, and a single poll could block for the full default (600s) if `timeout=None`. Now we set a deadline up front and cap each poll to whatever's left, with a hard ceiling of 60s per poll either way.
- A single dropped connection or read timeout during polling killed `wait()` outright, even though the job was still running fine. Transient errors now get retried as long as budget remains.
- `_update()` always sent `WAIT=-1`, so the server could block past our own read timeout and blow up with a `ReadTimeoutError`. `WAIT` is now bounded under the per-request timeout.
- Timeouts used to surface as a generic `DALServiceError` wrapping a raw `HTTPSConnectionPool(...): Read timed out` string. Added `DALJobTimeoutError` so callers can catch it directly.
- The backoff sleep between polls grew up to 120s regardless of remaining budget, so a short timeout could be missed by up to two minutes. It's now capped to whatever's left.

Tests cover timeout expiry, transient retry, and budget capping.


msdemlei (approved):
I was briefly wondering whether what this introduces actually is useful semantics for the timeout on wait or whether simply passing it through to the service as before is good enough.  But yes, given that there can be all kinds of failures in the infrastructure, adding local logic is the right thing to do.

I'm less convinced there is a need for MAX_JOB_POLL_BLOCK; why can't we just always pass remaining?  But if you've thought about it and figured it's what we want, so be it.

Thanks!

pyvo/dal/tap.py
ManonMarchand: Add `MAX_JOB_POLL_BLOCK` here

pyvo/dal/tap.py
```diff
@@ -1037,11 +1045,12 @@ def wait(self, *, phases=None, timeout=DEFAULT_JOB_WAIT_TIMEOUT):
         phases : list
             phases to wait for
         timeout : float or None
-            maximum time to wait in seconds. If None, defaults to
-            ``DEFAULT_JOB_WAIT_TIMEOUT``.
+            total budget in seconds; None waits indefinitely
```
ManonMarchand: Keep the sentence "defaults to ``DEFAULT_JOB_WAIT_TIMEOUT`". Also, maybe the true None behaviour might deserve a description in the changelog too

stvoutsin: Yes good catch, I've added both

stvoutsin:
> I was briefly wondering whether what this introduces actually is useful semantics for the timeout on wait or whether simply passing it through to the service as before is good enough. But yes, given that there can be all kinds of failures in the infrastructure, adding local logic is the right thing to do.
> 
> I'm less convinced there is a need for MAX_JOB_POLL_BLOCK; why can't we just always pass remaining? But if you've thought about it and figured it's what we want, so be it.
> 
> Thanks!

Thanks for the review and comments!

In terms of the last point, there's a couple of reasons for the use of that constant in the way the change was written (But do let me know if I've got this wrong):

`timeout=None` means remaining is infinite and we can't hand a socket an infinite timeout, so some finite cap has to exist.
Given that cap exists anyway using it for every poll (not just the None case) puts an upper bound on the failure detection latency. If we hold a connection open for the whole remaining budget and it dies silently we potentially don't get any feedback until the entire budget elapses.
Capping basically means a dead connection will surface faster and get retried on a fresh one (within about a minute) instead.

stvoutsin:
I've added a small correction to the sia2 docs to address a sphinx build error that was causing the CI step to fail. Let me know if you don't want this tagged along to this PR and I can do it in a separate one.

ManonMarchand (approved):

ManonMarchand:
I'll let Markus answer to the discussion on `MAX_JOB_POLL_BLOCK` but on my side this is ready to be merged. Thank you!

msdemlei:
Well, I´m satisfied that MAX_JOB_POLL_BLOCK is a good idea, and whether a minute is a reasonable choice we´ll see as we go.

